PR #245, opened 2026-04-25
[00009] Fix null Commits property in plan CLI commands

zachwolfe:
# Summary

## Changes

Fixed NullReferenceException in plan CLI commands by initializing all list properties to empty lists when they are null after YAML deserialization. Added test coverage to verify commands handle plans with null list properties.

## API Changes

None.

## Files Modified

- `src/Ivy.Tendril/Helpers/PlanCommandHelpers.cs` - Added null-coalescing initialization for all list properties in `ReadPlan` method
- `src/Ivy.Tendril.Test/PlanCliCommandTests.cs` - Added helper method `CreatePlanWithNullLists` and test case `PlanAddCommit_WithNullCommitsList_Succeeds`

## Commits

- dd61069 [00009] Fix null Commits property in plan CLI commands

### DIFF
--- a/src/Ivy.Tendril.Test/PlanCliCommandTests.cs
+++ b/src/Ivy.Tendril.Test/PlanCliCommandTests.cs
@@ -68,13 +68,14 @@ public class PlanCliCommandTests : IDisposable
 
     private string CreatePlanWithNullLists(string planId)
     {
-        var yaml = """
+        var yaml = $"""
             state: Draft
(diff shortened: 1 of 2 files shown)


zachwolfe:
@nielsbosma Tendril suggested this change about itself while I was doing something unrelated on the framework. I ran the new unit test with and without the other changes in this PR, and it seems valid (it throws without them).

PR #459, opened 2026-05-01
[JobsAppSheets]: fix datatable flicker on sheet open by stabilizing tree position

defymecobra:
## Problem

When opening a Sheet in the Jobs page (e.g. clicking on PlanId, LastOutput, or Prompt/Title), the DataTable would visually flicker — disappearing for a moment and reappearing with fresh state (scroll position and column widths lost).

### Root Cause

`RenderWithSheets` returned structurally different widget trees depending on whether a sheet was open:

```csharp
// No sheet — DataTable is a direct child of Layout
return layout | dataTable;                          // DataTable at Layout[0]

// With sheet — DataTable wrapped inside Fragment
return layout | new Fragment(dataTable, sheet);     // DataTable at Layout[0]/Fragment[0]
```

The Ivy Framework generates widget IDs from `TreePath` (position in the widget tree). When the DataTable moved from `Layout[0]` to `Layout[0]/Fragment[0]`, its ID changed. React treats a changed `key` as a completely different component — unmounting the old DataTable and mounting a new one, causing the flicker.

## Solution

Add the DataTable and Sheet as **separate direct children** of the Layout instead of wrapping them in a Fragment:

```csharp
layout |= dataTable;           // Always Layout[0] — stable TreePath
if (sheet != null)
    layout |= sheet;           // Layout[1] — appears/disappears independently
return layout;
```

The DataTable is always at `Layout[0]` regardless of sheet state, so its widget ID remains stable. React recognizes it as the same component and preserves its internal state. The Sheet is added/removed as a sibling at `Layout[1]` without affecting the DataTable's position.

## Changes

- **`JobsApp.Sheets.cs`** — Refactored `RenderWithSheets` to add DataTable and sheets as separate layout children instead of using `Fragment` wrapping. Converted early-return branches to `else if` chain with a single `return layout` at the end.

## Before

https://github.com/user-attachments/assets/5f078933-9bc1-478b-b01f-21f964fc07bb

## After

https://github.com/user-attachments/assets/b3b27969-68bd-4719-b162-d4ce803ba90f

### DIFF
--- a/src/Ivy.Tendril/Apps/JobsApp.Sheets.cs
+++ b/src/Ivy.Tendril/Apps/JobsApp.Sheets.cs
@@ -19,7 +19,7 @@ public partial class JobsApp
         IState<bool> hasStreamContent,
         LayoutView layout)
     {
-        layout |= dataTable;
+        object? activeSheet = null;
 
         if (showPlan.Value is { } planPath)
         {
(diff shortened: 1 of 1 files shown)


defymecobra:
Switched from direct layout children to `Fragment` wrapping as it's the semantically correct approach for this problem.

Instead of adding DataTable and Sheet as separate `layout |= ...` children, the output is now always wrapped in a single `Fragment(dataTable, activeSheet)`. This keeps DataTable at a stable `Fragment[0]` position regardless of sheet state, preventing the widget ID change that caused the flicker.

The `Fragment` constructor filters `null` children, so when no sheet is active it effectively becomes `Fragment(dataTable)` — the DataTable's tree position stays consistent.

PR #468, opened 2026-05-01
[DiffView]: hide duplicate header and add rename indicator

defymecobra:
## Problem

The Changes tab and commit detail sheet in Tendril displayed **two separate headers** for each file diff:

1. **Tendril header** — collapsible row with chevron, status icon, filename, and directory path
2. **DiffView internal header** — `OLD: full/path → NEW: full/path`

This caused visual duplication where the same filename and path appeared twice. Additionally, when a file was renamed (e.g. `MenuOrder.cs → Constants.cs`), the rename was not visible in the Tendril header — only the new name was shown.

**Before:**
```
▼ 📝 Constants.cs  src/Ivy.Tendril/Constants.cs        ← Tendril header
OLD: src/Ivy.Tendril/MenuOrder.cs → NEW: src/.../Constants.cs  ← DiffView header (duplicate)
```

**After:**
```
▼ 📝 MenuOrder.cs → Constants.cs  src/Ivy.Tendril/Constants.cs  ← single header with rename
```

### Hide DiffView header
- Used the new `ShowHeader(false)` prop on all `DiffView` instances in `ChangesTabView` and `PlanContentHelpers` to suppress the internal header that duplicated Tendril's own UI.

### Add rename detection
- Extended `FileDiff` record with an optional `OldFilePath` property.
- Updated `SplitDiffByFile` to extract the old path from `diff --git a/old b/new` headers and populate `OldFilePath` when the paths differ (indicating a rename).
- Updated file headers in both `ChangesTabView` and `PlanContentHelpers` to display `OldName → NewName` when a file is renamed.

## Changed files

| File | Change |
|------|--------|
| `PlanContentHelpers.cs` | Added `OldFilePath` to `FileDiff`, extract old path in `SplitDiffByFile`, rename header in commit detail sheet, `ShowHeader(false)` |
| `ChangesTabView.cs` | Rename header display, `ShowHeader(false)` |

## Testing

Verified in Tendril with a test plan using commit `0925b6d` (contains file renames `MenuOrder.cs → Constants.cs` and `DashboardStats.cs → DashboardModels.cs`):

- ✅ Changes tab — no duplicate headers, rename indicator visible
- ✅ Commit detail sheet — no duplicate headers, rename indicator visible
- ✅ Normal files — single header with filename and path
- ✅ New files — no `/dev/null` shown

image

### DIFF
--- a/src/Ivy.Tendril/Views/Tabs/ChangesTabView.cs
+++ b/src/Ivy.Tendril/Views/Tabs/ChangesTabView.cs
@@ -67,13 +67,26 @@ public class ChangesTabView(
             var chevronIcon = isExpanded ? Icons.ChevronDown : Icons.ChevronRight;
             var (statusIcon, statusColor) = PlanContentHelpers.GetFileStatusIconAndColor(fileDiff.Status);
             var fileName = Path.GetFileName(fileDiff.FilePath);
+            var isRenamed = fileDiff.OldFilePath != null;
+            var oldFileName = isRenamed ? Path.GetFileName(fileDiff.OldFilePath!) : null;
 
             var header = Layout.Horizontal()
                 .Gap(2)
                 | new Icon(chevronIcon).Small()
-                | new Icon(statusIcon).Small().Color(statusColor)
-                | Text.Block(fileName).Bold()
-                | Text.Muted(Path.GetDirectoryName(fileDiff.FilePath)?.Replace('\\', '/') ?? "");
+                | new Icon(statusIcon).Small().Color(statusColor);
+
+            if (isRenamed)
+            {
+                header |= Text.Block(oldFileName!).Bold();
+                header |= Text.Muted("→");
+                header |= Text.Block(fileName).Bold();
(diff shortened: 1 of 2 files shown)


defymecobra:
@claude review this

src/Ivy.Tendril/Views/Tabs/ChangesTabView.cs
```diff
@@ -91,7 +104,7 @@ public override object Build()
 
             if (isExpanded)
             {
-                diffsLayout |= new DiffView().Diff(fileDiff.Diff).Split().Width(Size.Full());
+                diffsLayout |= new DiffView().Diff(fileDiff.Diff).Split().Width(Size.Full()).ShowHeader(false);
```
rorychatt: I don't like this new API. See feedback in the correlated PR in Ivy-Framework

rorychatt (changes requested):

defymecobra:
Removed .ShowHeader(false) calls from ChangesTabView and PlanContentHelpers.
image
image

PR #481, opened 2026-05-05
Auto-install missing tools (PowerShell, gh, pandoc) from onboarding

dcrjodle:
## Summary
- Adds an "Install Now" button on the onboarding **Required Software** step for `pwsh`, `gh`, and `pandoc` so users don't have to leave the app to install them.
- Cross-OS strategy: `dotnet tool install --global PowerShell` (universal, no admin) for PowerShell; `brew install` on macOS / Linux-with-brew and `winget install` on Windows for `gh` and `pandoc`.
- Falls back to the existing "Install" URL link when auto-install isn't available (e.g. `git`, or Linux without brew).
- After a successful install the software check re-runs automatically; on failure the row flips to "Retry Install" and the error is rendered below the table.

## Restart caveat
PowerShell goes into `~/.dotnet/tools/` which is already on Tendril's PATH, so it's picked up immediately. brew/winget binaries normally are too, but if Tendril was launched without inheriting the user's shell PATH (Finder/Dock, or fresh winget PATH on Windows) the freshly installed tool may not be visible until restart. A muted hint at the bottom of the table tells the user to restart in that case.

## Test plan
- [ ] macOS: from a fresh `TENDRIL_HOME`, uninstall `pwsh` / `gh` / `pandoc` and confirm each "Install Now" button installs the tool and the row flips to ✅ Installed after the auto-recheck.
- [ ] macOS without Homebrew on PATH: confirm `gh` / `pandoc` rows show "Install Now" only when brew is resolvable at `/opt/homebrew/bin/brew` or `/usr/local/bin/brew`; otherwise fall back to the URL link.
- [ ] Windows: confirm `winget install` runs silently for `gh` and `pandoc` and the recheck succeeds.
- [ ] Linux without brew: confirm `gh` and `pandoc` rows still show the URL "Install" link (no auto-install button).
- [ ] Failure path: temporarily break the installer command, click "Install Now", confirm the row shows "Retry Install" and the error message appears below the table.
- [ ] Confirm `git` still falls back to the URL link on every OS (no auto-install attempted).

### DIFF
--- a/src/Ivy.Tendril/Apps/Onboarding/GitHubAuthenticator.cs
+++ b/src/Ivy.Tendril/Apps/Onboarding/GitHubAuthenticator.cs
@@ -1,0 +1,101 @@
+using System.Diagnostics;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Ivy.Tendril.Apps.Onboarding;
+
+internal static class GitHubAuthenticator
+{
+    public const string DeviceFlowUrl = "https://github.com/login/device";
+
+    private static readonly Regex CodePattern = new(
+        @"one-time code:\s+([A-Z0-9]+-[A-Z0-9]+)",
+        RegexOptions.IgnoreCase | RegexOptions.Compiled);
+
+    public static async Task<(bool Success, string Message)> AuthenticateAsync(
+        Action<string> onCodeDetected,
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #504, opened 2026-05-06
Replace --agent-cmd with --agent to use agent providers properly

nielsbosma:
## Summary
- Replaces the `--agent-cmd` hack (which overrode `psi.FileName` after provider built ProcessStartInfo) with `--agent` that selects a proper agent provider (`claude|gemini|codex|copilot|opencode`)
- Adds `agentOverride` parameter to `AgentProviderFactory.Resolve()` so the CLI flag flows through the provider abstraction
- Updates tests to use the new `Agent` property

## Test plan
- [x] Zero CS compilation errors
- [x] Dry-run tests updated to use `Agent = "claude"` instead of `AgentCmd = "nonexistent-binary"`


artem-ivy-ai:
### Staging removed

PR #506, opened 2026-05-06
[00160] Remove project name from verifications header in onboarding

dcrjodle:
Removed the project name from the onboarding verifications header. The text now always shows "Setting up verifications..." instead of conditionally appending the project name. Also removed the unused `currentProject` state and its associated set/clear calls.

## Files Modified

- **src/Ivy.Tendril/Apps/Onboarding/CompleteStepView.cs** — Removed `currentProject` state declaration, its usage in the loop and finally block, and simplified the `headerText` conditional.

---

Commits: ac0e3f4


artem-ivy-ai:
### Staging removed

PR #541, opened 2026-05-07
[00202] Fix onboarding loading indicator disappearing before JSON output renders

dcrjodle:
## Summary

Modified `NotifyingStream<T>.Write()` in `CompleteStepView.cs` to only dismiss the loading spinner when a visible event arrives. Previously, any stream write (including filtered `system` events) would hide the loading indicator, causing a ~2 second empty box before actual content rendered.

## Files Modified

- **src/Ivy.Tendril/Apps/Onboarding/CompleteStepView.cs** — Updated `Write()` method to check event type before triggering `_onFirstWrite()`

---

**Commits:**
- 46ff628 [00202] Fix loading indicator hiding on non-visible stream events

### DIFF
--- a/src/Ivy.Tendril/Apps/Onboarding/CompleteStepView.cs
+++ b/src/Ivy.Tendril/Apps/Onboarding/CompleteStepView.cs
@@ -174,10 +174,16 @@ public class CompleteStepView(
 
         public void Write(T data)
         {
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #542, opened 2026-05-07
[00204] Onboarding repo box path text should use primary foreground

dcrjodle:
# Summary

## Changes

Added explicit `.Color(Colors.Primary)` to the repo path label text in `ProjectRepoPickerView.cs` so it renders with primary foreground color instead of inheriting muted foreground from the parent Box's background.

## API Changes

None.

## Files Modified

- **src/Ivy.Tendril/Views/ProjectRepoPickerView.cs** — Added `.Color(Colors.Primary)` to path label Text.Block call (line 114)

---

Commits:
- 832ebba


artem-ivy-ai:
### Staging removed

PR #548, opened 2026-05-08
Add dotnet dev-certs trust step to installers

zachwolfe:
## Summary
- Adds a step to both the macOS (`install.sh`) and Windows (`install.ps1`) installers that trusts the .NET development HTTPS certificate
- Checks if the cert is already trusted before prompting, to avoid unnecessary user interaction


artem-ivy-ai:
### Staging deploy failed

> Branch does not exist

zachwolfe:
Note: Tested on a fresh macOS Tahoe VM. Has **_not_** been tested on Windows.

PR #559, opened 2026-05-09
Fix Windows command line length limit for agent prompts

nielsbosma:
Claude: pass prompt via stdin (-) instead of CLI argument (-- <prompt>).
Copilot: write prompt to temp file when PromptFilePath is set, since
copilot CLI doesn't support stdin input.
JobLauncher: write prompt to plan temp/ for non-stdin providers.


artem-ivy-ai:
### Staging removed

PR #562, opened 2026-05-10
Fix dev-mode shim: use dotnet exec instead of dotnet run

nielsbosma:
dotnet run triggers a build which fails when Tendril is already running
due to locked DLLs. Switch to dotnet exec on the already-built DLL to
avoid the build step entirely.


artem-ivy-ai:
### Staging removed

PR #565, opened 2026-05-10
Move [FORCE] from description string to firmware header (#565)

nielsbosma:
Pass -Force as a job arg instead of appending [FORCE] to the description.
JobLauncher extracts it into a Force: true firmware header value.
CreatePlan Program.md reads the header instead of parsing/stripping args.


artem-ivy-ai:
### Staging removed

Issue #571, opened 2026-05-11
Remove redundant add button in repository dialog


dcrjodle:
## Summary
An additional add button in the repository dialog is considered redundant and negatively impacts user experience. Users expect the repository to be added only when clicking the final Add button at the bottom of the dialog.

## Additional Context
- Slack thread: https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1778481705250119?thread_ts=1778481705.250119&cid=C0AS51ABBB6

## Reported by
Joshua Ng

---

### Attachments

![1778481705-250119-Screenshot 2026-05-11 at 08.40.29.png](https://stivytelemetry.blob.core.windows.net/dumpsterbot/4dbf293d-497d-4fa8-b1ef-751a3320e7d9/1778481705-250119-Screenshot_2026-05-11_at_08.40.29.png)

PR #579, opened 2026-05-11
Inline Memory/Tools file listings in firmware to avoid wasted tool calls (#571)

nielsbosma:


artem-ivy-ai:
### Staging removed

PR #581, opened 2026-05-11
Fix thread-safe log numbering, prompt file for non-stdin agents, consolidate raw log writing (#581)

nielsbosma:


artem-ivy-ai:
### Staging removed

PR #614, opened 2026-05-12
[Docs]: simplify hosted BasicAuth secrets in Setup

ArtemLazarchuk:
## Summary

Updates **Setup & Settings** docs with a short, layout-friendly section on enabling **browser login** when Tendril runs on a host that injects secrets as environment variables (e.g. Sliplane).

## What changed

- Documents **`TENDRIL_HOME`** plus **`BasicAuth__Users`**, **`BasicAuth__HashSecret`**, and **`BasicAuth__JwtSecret`**: what each is for, and that `JwtSecret` is an alternate env name for the same base64 value as `HashSecret`, not a separate secret.
- Mentions **`TENDRIL_AUTH_*`** as optional equivalents.
- Notes that the **REST API** uses **`api.apiKey`**, not these variables (with a link to the REST page).
- Replaces a wide three-column table with short paragraphs so the docs page does not overflow horizontally.

## Why

Operators need a **accurate, scannable** reference that matches what the platform UI shows (`BasicAuth__*`) without long unbroken lines in tables.

Знімок екрана 2026-05-12 о 12 31 13

### DIFF
--- a/src/Ivy.Tendril.Docs/Docs/03_Configuration/01_Setup.md
+++ b/src/Ivy.Tendril.Docs/Docs/03_Configuration/01_Setup.md
@@ -7,6 +7,12 @@ searchHints:
   - settings
   - projects
   - gui
+  - deployment
+  - docker
+  - secrets
+  - BasicAuth
+  - password
+  - hosted
 ---
 
 # Setup & Settings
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #621, opened 2026-05-12
Route promptware writes through CLI commands to bypass Windows glob permission bug (#29574)

nielsbosma:
Claude Code's Write(path/**) and Edit(path/**) glob patterns never match on Windows
in dontAsk mode. This routes all promptware file writes through scoped CLI commands
that match Bash(tendril*), which works correctly.

New CLI commands:
- tendril promptware run <name> [args] (renamed from tendril promptware <name>)
- tendril promptware write-memory <name> <file> (STDIN → Memory/<file>)
- tendril promptware write-tool <name> <file> (STDIN → Tools/<file>)
- tendril trash write <file> (STDIN → TENDRIL_HOME/Trash/<file>)

Permission changes:
- Removed Write(%PROMPTWARE_DIR%/**) and Edit(%PROMPTWARE_DIR%/**) from BaseTools
- Removed Write(%TENDRIL_HOME%/Trash/**) from CreatePlan extras
- Program.md is now immutable at runtime (agents write only via CLI)

Also removes TENDRIL_PROMPTWARE_DIR env var (doesn't propagate through Claude Code's
Bash tool to child processes).


artem-ivy-ai:
### Staging removed

PR #641, opened 2026-05-13
Fix JobDebugSheet context error by passing view instance instead of calling Build()

nielsbosma:


artem-ivy-ai:
### Staging removed

PR #642, opened 2026-05-13
Fix: "Open config.yaml" not opening when editor path is invalid or when accessed remotely

ArtemLazarchuk:
## Summary
- Capture the request host during `Build()` via `ConfigYamlUiHelper.CaptureHost` and pass it into `OpenOrNavigate` from click handlers, so loopback detection does not depend on `HttpContext` when the handler runs.
- Add an overload of `OpenOrNavigate` that takes `string? capturedHost`; rename loopback detection to `IsLoopbackHost`.
- Update `ConfigErrorApp` and `SettingsApp` to use the captured host.
- When launching the external editor from `PlatformHelper`, set `UseShellExecute = false` so the OS does not print a “file does not exist” style message to the terminal before .NET can handle the failure.

## Testing
- [ ] Manual: open Settings / config error UI, click “Edit config” / open config from localhost vs remote host (desktop shell vs browser) as applicable.


artem-ivy-ai:
### Staging removed

PR #646, opened 2026-05-14
Add Plan Id, CLI log link, and path dropdown menu to JobDebugSheet

nielsbosma:
- Show Plan Id field (extracted from PlanFile or reported/allocated IDs)
- Add Plan CLI Log field linking to the job-specific cli.jsonl
- Render permission denials in a CodeBlock instead of plain text
- Replace CopyToClipboard buttons with a dropdown menu offering both "Copy to Clipboard" and "Open in Editor"
- Extract shared logs directory lookup to reduce duplication


artem-ivy-ai:
### Staging removed

PR #652, opened 2026-05-14
Fix OutputSheet showing duplicate content on re-render

nielsbosma:
Capture initial OutputLines snapshot via UseState so JsonStream stays
frozen while only new lines arrive via the live Stream.


artem-ivy-ai:
### Staging removed

PR #653, opened 2026-05-14
Replace generic Args with named firmware headers and grant unrestricted Bash (#652)

nielsbosma:
- Rename Args → TaskDescription for CreatePlan firmware header
- Remove redundant Args for non-CreatePlan promptwares (already have TendrilPlanFolder)
- Grant unrestricted Bash to all promptwares (fixes non-deterministic heredoc permission denials)
- Simplify BuiltInExtraTools to only ExecutePlan → [Write, Edit]
- Add strict no-direct-fix policy to CreatePlan Program.md
- Update firmware template text to reference "named parameters"
- Remove TaskDescription from PathKeys (it's free text, not a path)


artem-ivy-ai:
### Staging removed

PR #654, opened 2026-05-14
Fix OutputSheet UseState crash by moving snapshot to JobsApp (#652)

nielsbosma:
OutputSheet.Build() is called directly from UseTrigger, not mounted as
a stateful child — UseState is unavailable there. Capture the initial
content snapshot in JobsApp where hooks are valid and pass it in.


artem-ivy-ai:
### Staging removed

PR #673, opened 2026-05-15
Fix premature plan visibility in Review/Drafts while jobs are still running

nielsbosma:
## Summary

- Hide plans from Review and Drafts apps when they have an active job (Running/Queued/Pending/Blocked)
- Correct PlanCountsService side menu counters to subtract plans with active jobs
- Self-heal premature state changes in JobCompletionHandler by always overriding on completion
- Strengthen ExecutePlan Program.md to explicitly forbid manual state changes

## Test plan

- [x] All 1429 unit tests pass
- [x] Build succeeds


artem-ivy-ai:
### Staging removed

PR #679, opened 2026-05-15
Fix Review tab refresh stuck in Loading state

nielsbosma:
## Summary
- Use `Revalidate()` instead of full remount so existing content stays visible while data refreshes in background
- Change query key from empty string to `null` for idle state when no plan is selected
- Fix GitService timeout pattern: `ReadToEndAsync()` before `WaitForExit()` so timeouts actually fire instead of blocking forever

## Test plan
- [x] All 36 GitService tests pass
- [x] All 18 PlanContentHelper tests pass
- [ ] Manual: open Review app, select a plan, click tab refresh — content stays visible, refreshes in background
- [ ] Manual: verify initial load of a plan still shows "Loading..." correctly

### DIFF
--- a/src/Ivy.Tendril/Apps/Review/ContentView.cs
+++ b/src/Ivy.Tendril/Apps/Review/ContentView.cs
@@ -1,3 +1,4 @@
+using System.Reactive.Disposables;
 using Ivy.Core;
 using Ivy.Tendril.Apps.Plans;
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging deploy failed

> Github repository not found. It might be deleted or you are missing access to it.

PR #680, opened 2026-05-15
Fix Reset to Draft not deleting the worktrees directory

nielsbosma:
RemoveWorktrees unregisters git worktrees and deletes subdirectories
but leaves the parent worktrees/ folder behind. Add explicit deletion
of the directory after cleanup, matching CleanupPlanWorktrees behavior.


artem-ivy-ai:
### Staging removed

PR #688, opened 2026-05-15
Fix OutputSheet context error and remove autofocus from import dialog

nielsbosma:
## Summary
- Pass OutputSheet instance to Sheet instead of calling .Build() manually, fixing InvalidOperationException when clicking Agent Output
- Remove AutoFocus from repository select in ImportIssuesDialog


artem-ivy-ai:
### Staging removed

PR #693, opened 2026-05-15
Remove auth credentials from committed team config

nielsbosma:
Remove hardcoded auth block (password hash + zero-byte hashSecret) from TeamIvyConfig. Auth should come from environment variables instead.


artem-ivy-ai:
### Staging removed

PR #694, opened 2026-05-15
Fix duplicate content in Agent Output during live streaming

nielsbosma:
Freeze the JsonStream snapshot at mount time using UseRef so it doesn't
grow to include lines already delivered via the live Stream subscription.


artem-ivy-ai:
### Staging removed

PR #704, opened 2026-05-17
Add OS version, app version, and agent to PostHog telemetry

nielsbosma:
Adds three new dimensions to PostHog analytics:

- **OS version** — `os_version` as both person property and super property (e.g. "Microsoft Windows NT 10.0.26100.0")
- **App version** — `app_version` as super property on every event (was only on identify + app_started)
- **Agent provider** — `agent` property on `job_completed`, `plan_created`, and `pr_created` events (e.g. "claude", "codex", "gemini")

All additions comply with the telemetry data classification policy (non-identifying, aggregate-useful values).


artem-ivy-ai:
### Staging removed

PR #720, opened 2026-05-18
Fix RetryPlan crash: stream raw log to disk, simplify promptware

nielsbosma:
- Stream agent output to .raw.jsonl as it arrives (AutoFlush StreamWriter)
  so crash/kill no longer loses all output data
- Rewrite RetryPlan Program.md to work in existing worktree instead of
  creating/destroying worktrees (which caused the taskkill crash)
- Skip redundant WriteRawLog if streaming already populated the file
- Close raw log writer before completion handler checks file


artem-ivy-ai:
### Staging removed

PR #733, opened 2026-05-18
Fix: store full commit hashes in syncer and refresh UI after sync

nielsbosma:
## Summary
- `PlanArtifactSyncer.ExtractCommitsFromWorktree` now stores full 40-char SHAs instead of truncating to 9
- `AddCommitsToPlan` uses prefix matching for dedup (handles mixed full/abbreviated hashes)
- `SynchronizeWorktreeAsync` refreshes `selectedPlanState` after artifact sync so the Git tab re-renders with updated commits

## Test plan
- [x] All 1434 tests pass
- [x] Build succeeds
- [x] Plan 21 yaml manually corrected to full hashes

### DIFF
--- a/src/Ivy.Tendril/Apps/Review/ContentView.cs
+++ b/src/Ivy.Tendril/Apps/Review/ContentView.cs
@@ -431,7 +431,7 @@ public class ContentView(
                     return null!;
                 },
                 syncingWorktrees.Value,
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #735, opened 2026-05-18
Fix review action commands to work with RunPowerShellAction

nielsbosma:
The commands are passed via pwsh -Command, so they should be PowerShell
expressions (& script.ps1), not full pwsh invocations. Also fix example
config to use 'command' instead of 'action' to match the C# model.


artem-ivy-ai:
### Staging removed

PR #736, opened 2026-05-18
Fix review action command paths to use quoted PowerShell syntax

nielsbosma:
The & call operator needs the path quoted with single quotes and
prefixed with .\ so PowerShell treats it as a file path rather than
trying to load it as a module.


artem-ivy-ai:
### Staging removed

PR #738, opened 2026-05-18
Remove PlanEvaluator hooks that open Tendril.ps1 in VS Code

nielsbosma:
Start-Process -FilePath 'tendril' resolves to Tendril.ps1 in PATH, which Windows opens with its default handler (VS Code) instead of executing it. Removed from all three projects.


artem-ivy-ai:
### Staging removed

PR #748, opened 2026-05-18
Fix Changes tab blinking every 30s during background revalidation

nielsbosma:
## Summary
- Only show "Loading..." in ChangesTabView when there is no existing data (`changesData is null`)
- Prevents the 30-second poll timer from replacing rendered diffs with loading text during background revalidation


artem-ivy-ai:
### Staging removed

PR #791, opened 2026-05-20
fix: make process view targets import conditional with nuget fallback

rorychatt:
Resolves the MSB4019 targets import error when packing Tendril in CI where the sibling framework repository is not cloned. Uses sibling framework targets if present, falling back to NuGet package targets.


rorychatt:
@claude review this

artem-ivy-ai:
### Staging removed

PR #839, opened 2026-05-23
[ProjectInputStepView]: show "Cancel" instead of "Back" on first dialog step

### LINKED ISSUE #823: Dialog back button should be cancel when there's nothing to back to
## Summary
The "Add Project" dialog displays a back button at the bottom left, but this is the first step in the flow with nothing to navigate back to. The button should either be labeled "Cancel" or removed entirely.

## Expected vs Actual Behavior
**Actual:** Dialog shows a "Back" button when there is no previous step to return to.

**Expected:** The button should either:
- Be labeled "Cancel" to accurately reflect its function, or
- Be removed from the dialog

## Additional Context
The issue is visible in the "Add Project" dialog where repositories are being configured. The back button appears alongside "Manual Setup" and "AI Setup" options at the bottom of the dialog.

---

[Slack thread](https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1779367567246919?thread_ts=1779367567.246919&cid=C0AS51ABBB6)

Reported by @dcrjodle

---

defymecobra:
## Problem

The `ProjectInputStepView` component is used in two different contexts:

| Context | File | Step Position | Back Action |
|---------|------|---------------|-------------|
| **OnboardingApp** | `OnboardingApp.cs` | 3rd step of 4 | Navigates to previous step |
| **AddProjectDialog** | `AddProjectDialog.cs` | **1st step of 3** | Closes the dialog (`CancelAndClose()`) |

In `AddProjectDialog`, the view is the **first step** in the flow — there is no previous step to navigate back to. However, the button was hardcoded as **"← Back"**, which is misleading. The `onBack` action actually calls `CancelAndClose()`, so the button should reflect that.

## What needed to be done

The button label and icon needed to adapt to the context:
- When used as a **middle/later step** (e.g. in Onboarding) → show **"← Back"** with an arrow icon
- When used as the **first step** (e.g. in AddProjectDialog) → show **"Cancel"** without an arrow icon

### `ProjectInputStepView.cs`

- Added a new optional constructor parameter `backButtonText` (default: `"Back"`) — following the existing pattern already used for `skipButtonText` and `nextButtonText`
- The button now uses `backButtonText` instead of the hardcoded `"Back"` string
- The arrow icon (`Icons.ArrowLeft`) is now **conditionally rendered** — it only appears when the button text is `"Back"`, and is hidden for `"Cancel"` since a left arrow doesn't make sense for a cancel action

### `AddProjectDialog.cs`

- Passed `backButtonText: "Cancel"` when constructing `ProjectInputStepView` at step 0

### No changes to `OnboardingApp.cs`

The default value `"Back"` preserves the existing behavior — no modifications needed.

## Result

| Context | Before | After |
|---------|--------|-------|
| AddProjectDialog (step 0) | ← Back | Cancel |
| OnboardingApp (step 2) | ← Back | ← Back (unchanged) |

image
image

### DIFF
--- a/src/Ivy.Tendril/Apps/Onboarding/ProjectInputStepView.cs
+++ b/src/Ivy.Tendril/Apps/Onboarding/ProjectInputStepView.cs
@@ -8,12 +8,11 @@ public class ProjectInputStepView(
     IState<List<RepoRef>> selectedRepos,
     IState<string> projectName,
     IState<bool> isStepLoading,
-    Action onBack,
     Action onNext,
+    Action? onBack = null,
     Action? onSkip = null,
     string skipButtonText = "Skip",
     string nextButtonText = "AI Setup",
-    string backButtonText = "Back",
     string title = "Setup your first project",
     bool disableSkipWhenCannotContinue = false) : ViewBase
 {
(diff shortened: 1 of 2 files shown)


defymecobra:
image
Made the onBack parameter optional in ProjectInputStepView. When omitted (e.g. on the first step of AddProjectDialog), the back button is hidden since the dialog's close button already handles dismissal.

artem-ivy-ai:
### Staging removed

PR #841, opened 2026-05-23
[Projects]: prevent adding invalid local repositories to projects

### LINKED ISSUE #821: Projects can be created with repos that have failed validation
## Summary
Users are able to create and add a project with repositories that have failed validation.

## Additional Context
This appears to be a validation bypass issue where the system should prevent project creation when associated repositories fail validation checks.

---

[Slack thread](https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1779367400882599?thread_ts=1779367400.882599&cid=C0AS51ABBB6)

Reported by @dcrjodle

defymecobra:
## What was the issue?
Users were able to create and save projects with local repository paths that either did not exist or were not valid git repositories. While the UI visually flagged these invalid paths with a warning icon and red text, it did not block the user from adding them to the project or saving the configuration. This allowed invalid state to be persisted to `config.yaml`, leading to downstream errors (such as exceptions thrown by `PlanValidationService` when attempting to create a plan for that project).

## What needed to be done?
Validation for local repository paths needed to be enforced at the UI level *before* they are added to a project's repository list, and the UI needed to prevent saving any project that contains invalid repository paths.

## What I did
1. **Centralized Validation:** Added `ValidateLocalRepo` and `HasInvalidLocalRepos` methods to `RepoPathValidator.cs`. This encapsulates the logic for checking if a directory exists and contains a `.git` folder, replacing the scattered, UI-only checks.
2. **Blocked Addition of Invalid Paths:** Updated `ProjectRepoPickerView.AddAsync()`. Now, if a user attempts to add an invalid local repository path, the UI immediately shows an error message and prevents the path from being added to the list.
3. **Disabled Saving with Invalid State:** Updated `EditProjectDialog.cs` to disable the "Save" (or "Add") button if the project currently contains any invalid local repositories. A tooltip was added to explain why the button is disabled.

## Why?
- Rejecting invalid paths at the input stage (`ProjectRepoPickerView`) is the most robust way to prevent bad data entry.
- Disabling the "Save" button in `EditProjectDialog` acts as a crucial safety net for users who might open an *already existing* project configuration that contains invalid paths from before this fix was applied.
- By keeping the validation purely at the UI level (rather than strictly blocking it deep in `ConfigService.SaveSettings()`), we avoid the risk of breaking existing installations or causing false positives during background operations, maintaining a minimal and safe diff.

image

### DIFF
--- a/src/Ivy.Tendril/Apps/Setup/Dialogs/EditProjectDialog.cs
+++ b/src/Ivy.Tendril/Apps/Setup/Dialogs/EditProjectDialog.cs
@@ -135,6 +135,9 @@ public class EditProjectDialog(
                                        : new Spacer());
         }
 
+        var tendrilHome = Environment.GetEnvironmentVariable("TENDRIL_HOME");
+        var hasInvalidRepos = RepoPathValidator.HasInvalidLocalRepos(editRepos.Value, tendrilHome);
+
         var mainDialog = new Dialog(
             _ => _editIndex.Set(-1),
             new DialogHeader(isNew ? "Add Project" : $"Edit Project: {editName.Value}"),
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #846, opened 2026-05-24
fix(opencode): suppress intermediate step_finish events in output view

nielsbosma:
OpenCode emits step_finish after every LLM turn with reason "tool-calls".
These were being emitted as ResultEvents with is_success=false, causing
the AgentOutputView to render "X Error" blocks between each tool call.

Now only terminal step_finish events (reason "stop" or "error") produce
a ResultEvent. Intermediate steps are suppressed.


artem-ivy-ai:
### Staging removed

PR #855, opened 2026-05-25
fix: prepend .NET 10 to PATH in install.sh

rorychatt:
This PR fixes a bug in the macOS/Linux installer script where .NET 10 is appended instead of prepended to PATH. Prepending it ensures the correct .NET 10 SDK is used during the installer's execution instead of any older pre-existing .NET SDK versions.


rorychatt:
@claude review this

artem-ivy-ai:
### Staging removed

PR #879, opened 2026-05-26
feat: open CreatePlanDialog from wallpaper New Plan button

nielsbosma:
Previously the OnCreate handler just navigated to DraftsApp.
Now it uses CreatePlanDialogLauncher so users can immediately
start creating a plan from the home screen.


artem-ivy-ai:
### Staging removed

PR #880, opened 2026-05-26
fix: handle null models during agent switch and remove alias from label

nielsbosma:
## Summary
- Guard against null `modelsQuery.Value` with `?? []` to prevent `ArgumentNullException` when switching agents
- Show loading state on model dropdowns while fetching catalog
- Use `DisplayName` only (no alias) for model option labels


artem-ivy-ai:
### Staging removed

PR #884, opened 2026-05-26
fix: wait for tunnel to become routable before declaring connected

nielsbosma:
The TunnelConnected event fired as soon as cloudflared printed the URL,
but Cloudflare's edge network needs 15-30s to route traffic. This caused
502 Bad Gateway errors despite the UI showing "Tunnel Active".

- Add health-check polling in CloudflaredService that waits for non-502
  response before firing TunnelConnected
- Add loading UI state with spinner while tunnel is starting
- Add "Copy URL" button for easy clipboard access
- Add E2E tests verifying actual traffic proxying through the tunnel


artem-ivy-ai:
### Staging removed

PR #882, opened 2026-05-26
fix: make process view buttons equal width and change pulse to gray

nielsbosma:

PR #885, opened 2026-05-26
fix: use "default" sentinel value so model dropdown shows Default option (#882)

nielsbosma:
## Summary
- The Ivy Framework's `SelectInput` filters out options with empty string values, making the "Default" option invisible in Model Per Profile dropdowns
- Changed the option value from `""` to `"default"` which `AgentProviderFactory` already treats as no-override
- Updated `GetProfileModel` to return `"default"` instead of `""` for unconfigured profiles so the dropdown shows the correct selection

## Test plan
- [x] All 1299 tests pass
- [x] Build succeeds with no warnings


artem-ivy-ai:
### Staging removed

PR #889, opened 2026-05-27
fix: use semantic color tokens in TendrilProcessView for dark mode support

nielsbosma:
Replace hardcoded rgba values and primary color tints with semantic CSS
variables (--color-card, --color-border, --color-foreground) so the
widget adapts to both light and dark themes.


artem-ivy-ai:
### Staging removed

PR #893, opened 2026-05-27
fix: resolve cloudflared 502 by reading actual server address from IServer

nielsbosma:
## Summary

- Cloudflared tunnel was always getting 502 because the origin URL defaulted to `http://localhost:5010` while the Ivy framework binds to `https://localhost:5010`
- Root cause: `app.Urls.FirstOrDefault()` was called during build time (before server started), so `TENDRIL_URL` was never set
- Fix: inject `IServer` into `CloudflaredService` and read `IServerAddressesFeature` after `ApplicationStarted` fires
- Also normalizes `localhost` → `127.0.0.1` to avoid IPv4/IPv6 ambiguity on Windows

## Test plan

- [x] Build succeeds (0 errors, 0 warnings)
- [x] 2056 unit tests pass
- [ ] Verify tunnel connects successfully with `TENDRIL_VERBOSE=1` and confirm log shows `https://127.0.0.1:5010` as origin URL


artem-ivy-ai:
### Staging removed

PR #894, opened 2026-05-27
fix: switch copilot provider to stdin prompt to avoid Win32 command-line length limit

nielsbosma:
## Summary
- Switch CopilotCli from passing prompt as `-p` argument to piping via stdin, avoiding Windows' ~32K command-line length limit (Win32 error 206)
- Improve Win32Exception error messages in JobLauncher to differentiate "binary not found" (code 2) from "command line too long" (code 206)
- Update tests to reflect new stdin transport behavior

## Test plan
- [x] All CopilotCli tests pass (45/45)
- [ ] Run a copilot job and verify it launches without error 206


artem-ivy-ai:
### Staging removed

PR #895, opened 2026-05-27
feat: add Test Agent button to settings for install/auth/model checks

nielsbosma:
Adds a "Test Agent" button to Settings > Coding Agent that runs the same
health checks as onboarding: installation, authentication, and model
validation. Models are deduplicated so the same model is only tested once
even if used across multiple profiles.


artem-ivy-ai:
### Staging removed

PR #898, opened 2026-05-27
feat: simplify model catalog, extract agent test dialog, resolve default models

nielsbosma:
- Simplify Claude model catalog to alias-based IDs (opus, sonnet, haiku) for Bedrock compatibility
- Extract test dialog into AgentTestDialog ViewBase class following project dialog patterns
- Resolve default models from agent descriptor when profiles are set to "Default"
- Add Bedrock/Vertex provider detection in health check auth
- Remove unused Alias/Hidden properties from ModelInfo
- Fix Cancel button reactivity by passing IState<bool> instead of captured bool


artem-ivy-ai:
### Staging removed

PR #905, opened 2026-05-27
fix: use default models for Copilot profiles in new installations

nielsbosma:


artem-ivy-ai:
### Staging removed

PR #909, opened 2026-05-27
fix: write raw agent output to .raw.jsonl instead of EventWire format

nielsbosma:
## Summary
- `.raw.jsonl` now captures raw stdout lines from the agent process instead of parsed EventWire events
- Moved `AppendToRawLog` call to before parsing in `EnqueueOutput` so it writes the original line
- Removed redundant `WriteRawLog` call from `JobCompletionHandler` (streaming writer already captures during execution)


artem-ivy-ai:
### Staging removed

PR #912, opened 2026-05-27
fix: avoid command-line length limit when launching antigravity

nielsbosma:
Switch AntigravityCli to PromptTransport.File and launch via pwsh to
read the prompt from a temp file, bypassing Windows' 32K CreateProcess
command-line limit.


artem-ivy-ai:
### Staging deploy failed

> Internal server error

PR #913, opened 2026-05-27
fix: clean up doctor output — version parsing, model validation, formatting

nielsbosma:
- Extract semver from verbose --version output (Claude, Codex, Copilot)
- Treat "default" model as no-model and still run a ping to validate
- Add live validation for Antigravity and OpenCode default models
- Cache model validation results to avoid redundant checks per agent
- Remove leading whitespace from doctor status lines
- Remove unregistered gemini agent from config


artem-ivy-ai:
### Staging removed

PR #921, opened 2026-05-28
fix: resolve release workflow and build warnings

rorychatt:
This PR removes the upload-install-scripts job, configures actions to run with Node 24, and fixes the Assembly.Location single-file compiler warning.


artem-ivy-ai:
### Staging removed

PR #923, opened 2026-05-28
fix: use clean console formatter for CLI output

nielsbosma:
Replaces the default .NET console logger (which outputs verbose structured logging with category names and event IDs) with a custom CleanConsoleFormatter that only prints the message.


artem-ivy-ai:
### Staging removed

PR #936, opened 2026-05-28
fix: restore malware scan and set max-parallel: 1

rorychatt:
Restores malware_block: true and adds max-parallel: 1 to the desktop publish strategy to prevent parallel jobs from colliding on the same TOTP credential during code signing.


artem-ivy-ai:
### Staging removed

PR #940, opened 2026-05-28
fix: swap updatingPlansCount and executingPlansCount in process view

nielsbosma:
The loop arrow on Drafts should show updating plans (revising stays in Drafts) and the forward arrow to Review should show executing plans (executing advances to Review).


artem-ivy-ai:
### Staging removed

PR #947, opened 2026-05-28
feat: add description field to ToolCallWire for human-readable tool titles

nielsbosma:
Extract the description field from tool inputs (when available) and surface
it as a top-level field on the tool_call EventWire event. The frontend now
prefers this description for the tool call title, falling back to command,
file_path, path, or pattern. Implemented for Claude, Copilot, and OpenCode
parsers.


artem-ivy-ai:
### Staging removed

PR #977, opened 2026-05-29
fix: promptware memory/tools written to deployed copy instead of source

nielsbosma:
## Summary

- Removes the `IsEmbeddedAvailable()` guard in `ResolvePromptsRoot`
- Source Promptwares folder is now always preferred when it exists on disk
- Fixes memory/tool writes landing in `TeamIvyConfig/Promptwares/` instead of `src/Ivy.Tendril/Promptwares/`

## Test plan

- [x] Build succeeds
- [x] 1302 unit tests pass
- [ ] Run a job that writes memory, verify it goes to `src/Ivy.Tendril/Promptwares/<type>/Memory/`


artem-ivy-ai:
### Staging removed

PR #996, opened 2026-05-30
[ProjectRepoPicker]: style folder select as Ghost suffix

defymecobra:
## Problem

The "Browse" button in the Add Project dialog (`ProjectRepoPickerView`) was using the `.Outline()` variant while being placed inside a `.Suffix()` slot of the repository path `TextInput`. This created several visual issues:

1. **Gray padding strips** — The suffix container applies its own `px-3` padding and `bg-muted` background. Since the `.Outline()` button renders its own border and background, visible gray strips appeared between the button edges and the container boundaries.
2. **Double borders** — The `.Outline()` button drew its own border on all four sides, while the `TextInput` wrapper also drew an outer border around the entire component (including the suffix). This resulted in overlapping borders that looked unpolished.
3. **Muted button text** — The button label and icon inherited the `text-muted-foreground` color from the affix container, making the button appear disabled.

## What needed to be done

Make the "Browse" button visually integrate with the `TextInput` as a proper "attix" — an attached element that looks like a seamless extension of the input field, matching the design reference in the issue.

## What was changed and why

Changed the button variant from `.Outline()` to `.Ghost()` in `ProjectRepoPickerView.cs`.

The `.Suffix()` slot in Ivy's `TextInput` is designed so that the **affix container itself** acts as the visual button surface — it provides the muted background, rounded corners on the outer side, and a clear boundary against the input field. All existing suffix buttons across the framework (in `TextInputApp`, `SelectInputApp`, `ColorInputApp`, etc.) use `.Ghost().Small()` for this reason.

By switching to `.Ghost()`, the button becomes transparent and lets the suffix container's styling define the visual appearance, which is exactly the intended attix pattern. The button's own border and background no longer conflict with the container's styling.

A companion change in `Ivy-Framework` (`[&_button]:text-foreground` in `affixEmbeddedButtonClasses`) ensures the button text renders in the standard foreground color instead of the muted gray inherited from the container.

### Changed file

- `src/Ivy.Tendril/Apps/Views/ProjectRepoPickerView.cs`

image
image

### DIFF
--- a/src/Ivy.Tendril/Apps/Views/ProjectRepoPickerView.cs
+++ b/src/Ivy.Tendril/Apps/Views/ProjectRepoPickerView.cs
@@ -81,22 +81,20 @@ public class ProjectRepoPickerView(
         object pickerControls;
         if (isDesktop)
         {
-            pickerControls = Layout.Horizontal().Gap(2).Width(Size.Full())
-                             | inputValue.ToTextInput("Repository URL or Local Path")
+            pickerControls = inputValue.ToTextInput("Repository URL or Local Path")
                                  .Width(Size.Grow())
                                  .OnSubmit(() => { _ = AddAsync(); })
-                             | new Button("Browse").Icon(Icons.FolderOpen).Outline()
-                                 .OnClick(() =>
-                                 {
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #1015, opened 2026-06-02
ci: fix Linux ARM64 pack command and add version to upload-artifact name

rorychatt:
This PR fixes the `--noPortable` error on Linux ARM64 (by checking `matrix.rid` instead of `matrix.os`), uses robust RID-based checks for other packing/signing steps, and adds the resolved version to the upload-artifact name to ensure artifacts contain the version in their filenames when downloaded.


artem-ivy-ai:
### Staging removed

PR #1071, opened 2026-06-03
[00093] Auto-select first plan when opening Icebox

nielsbosma:
Added auto-selection logic to IceboxApp to automatically select the first plan when the Icebox tab is opened and no plan is currently selected. This matches the existing behavior in DraftsApp.

## API Changes

None.

## Files Modified

- **Apps/IceboxApp.cs** — Added auto-selection of first plan when none is selected

---

**Commits:**
- 50e7e1d Auto-select first plan when opening Icebox


artem-ivy-ai:
### Staging removed

PR #1080, opened 2026-06-03
Fix plan.yaml lost-write race condition on recommendation accept

nielsbosma:
The Accept button in Review fired three concurrent background writes
(ResetVerificationsForRetry, TransitionState, UpdateRecommendationState)
that each did independent read-modify-write on plan.yaml. The last
writer won, silently dropping the others' changes — causing accepted/
declined recommendations to revert to Pending.

Fixes:
- Serialize all in-process plan.yaml writes through a SemaphoreSlim
- Add PlanFileLock for cross-process coordination with CLI commands
- Add AcceptRecommendationAndRetry that performs all three mutations
  in a single atomic read-modify-write
- Read recommendations from database (not stale file) in Review UI
- Remove recommendation generation from RetryPlan (not its job)


artem-ivy-ai:
### Staging removed

PR #1090, opened 2026-06-04
Fix race condition in WaitForJobs unblocking and duplicate job detection

nielsbosma:
- Fix HandleWaitForJobsDependents: remove blocked job from dictionary
  before calling startJobSkipDepCheck to prevent conflict check from
  seeing the placeholder as an active job
- Add Blocked status to TryRejectConflictingJob so WaitForJobs-blocked
  ExecutePlan jobs are correctly detected as conflicts
- Make JobStatusCommand resilient to HTTP failures (log instead of crash)
- Fix broken DoctorChecksTests.PrintStatus test (correct types/visibility)
- Add test for Blocked-state conflict detection


artem-ivy-ai:
### Staging removed

PR #1107, opened 2026-06-08
Fix job cost calculation: run for all statuses, use inline ResultEvent data

nielsbosma:
Previously cost calculation only ran for Completed jobs. Failed/stopped/timeout
jobs still consume tokens but never got costs populated. Also extracts cost from
the inline ResultEvent (already parsed from stream) as the primary path, avoiding
the 30s delay and session-file discovery issues.


artem-ivy-ai:
### Staging removed

PR #1115, opened 2026-06-08
Fix IvyFrameworkVerification leaving zombie processes after test runs

nielsbosma:
- Fix forward-slash path in PowerShell -like filter (never matched on Windows)
- Replace one-server-per-spec architecture with single shared server via
  Playwright globalSetup/globalTeardown (eliminates 15 zombie processes)
- Use shell: false in dotnet spawn to avoid unkillable cmd.exe process trees
- Replace silent post-test cleanup with 3-step escalation that verifies success


artem-ivy-ai:
### Staging removed

PR #1128, opened 2026-06-09
Fix Synchronize button not refreshing commits table

nielsbosma:
Read plan from disk after SyncPlanArtifacts instead of from the database, which hasn't been synced yet due to the 500ms debounce on PlanWatcherService.

Adds `GetPlanByFolderFromDisk` to `IPlanReaderService` (with default interface implementation) so the Synchronize action reads fresh commit data immediately from the filesystem rather than the stale database cache.


artem-ivy-ai:
### Staging removed

PR #1129, opened 2026-06-09
Add PlanMarkdownView widget with pinned fixed-content slot

dcrjodle:
## Summary

Introduces an external **`PlanMarkdownView`** widget in `Ivy.Tendril.Widgets` that renders plan markdown in its own scroll container alongside a **`FixedContent` slot** for interactive elements that should stay pinned in place while the plan content scrolls. Wires it into the Drafts **Plan** tab with a placeholder pinned element.

## What's included

- **`Widgets/PlanMarkdownView.cs`** — external widget with `Content`, `Article`, `DangerouslyAllowLocalFiles`, `OnLinkClick`, and a `[Slot("FixedContent")]` exposed via a `.FixedContent(...)` builder. Follows the `AgentOutputView` / `TendrilProcessView` conventions (uses the framework's `EventHandler<Event<T,V>>?` pattern for the event prop).
- **`frontend/src/PlanMarkdownView.tsx`** + **`plan-markdown-view.css`** — renders markdown via `react-markdown` / `remark-gfm` / `rehype-highlight` inside a scrolling body, with the `FixedContent` slot as a flex sibling **outside** the scroll region so it stays pinned. Registered in `index.ts`.
- **Typography** — replicates the Ivy framework Markdown widget's `typography` map and `article` variant (heading top-margins, h2 divider, relaxed body line-height) as scoped CSS on `.pmv-markdown`, using the host design-system CSS variables. The widget bundle's Tailwind build doesn't emit utility classes for markdown content, so this matches the existing `AgentOutputView` hand-written-CSS convention.
- **Wiring** (`Apps/Drafts/PlanTabView.cs`, `Apps/Drafts/ContentView.cs`) — the Plan tab uses `PlanMarkdownView` with a placeholder fixed `Card`. The Plan tab no longer wraps content in `Cap()` (whose outer scroll would also scroll the pinned element); the widget reproduces `Cap()`'s **1.5rem left inset** + 200-unit max-width so the markdown lines up exactly with the **Details** tab content. The edit-mode branch is made self-contained (own scroll + inset) since `Cap()` is gone.

## Notes

- The fixed-slot element is a **placeholder** (`Actions` card with a button) — intended to be replaced with the real interactive content.
- Like `AgentOutputView`, fenced code gets a styled block (background/border/mono) but not per-token syntax colors; the framework's richer `CodeBlock` was intentionally not ported.

## Testing

- `Ivy.Tendril` and `Ivy.Tendril.Widgets` build clean against local framework source (`-p:IvySource=true`).
- Widget frontend bundle builds clean (tsc + vite); `.pmv-markdown` / `.pmv-article` rules verified present in the compiled CSS.

### DIFF
--- a/src/Ivy.Tendril.Widgets/Widgets/PlanMarkdownView.cs
+++ b/src/Ivy.Tendril.Widgets/Widgets/PlanMarkdownView.cs
@@ -1,0 +1,74 @@
+using Ivy;
+using Ivy.Core;
+using Ivy.Core.ExternalWidgets;
+
+namespace Ivy.Widgets.PlanMarkdownView;
+
+/// <summary>
+/// Renders plan markdown in its own internal scroll container, alongside a
+/// <c>FixedContent</c> slot that is pinned in place and unaffected by the
+/// markdown scroll. Use the slot for interactive elements that should stay put
+/// (to the right of the markdown) while the plan content scrolls.
+/// </summary>
+[ExternalWidget(
+    "Widgets/frontend/dist/ivy-tendril-widgets.js",
+    StylePath = "Widgets/frontend/dist/ivy-tendril-widgets.css",
+    ExportName = "PlanMarkdownView",
+    GlobalName = "IvyTendrilWidgets"
+)]
+[Slot("FixedContent")]
+public record PlanMarkdownView : WidgetBase<PlanMarkdownView>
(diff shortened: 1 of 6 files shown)


artem-ivy-ai:
### Staging removed

PR #1131, opened 2026-06-09
Deduplicate SyncRepo jobs for the same repository

nielsbosma:
## Summary
- Detect when a SyncRepo is already active (Running/Queued/Pending/Blocked) for the same RepoPath
- Return the existing job ID instead of spawning a duplicate
- Callers that use the ID for WaitForJobs work seamlessly with the shared job

## Test plan
- [x] Build succeeds
- [x] All 1369 tests pass


artem-ivy-ai:
### Staging removed

PR #1137, opened 2026-06-09
Preserve agent status message when CreatePlan fails to produce a plan

nielsbosma:
## Summary

- Move StatusMessage clearing to after HandleCompletion, only for jobs that stay Completed
- In MarkCreatePlanFailed, fall back to agent's last status message before "No plan created"
- Update CreatePlan promptware to always send a descriptive final status before exiting without a plan

Fixes the issue where CreatePlan jobs that legitimately decide not to create a plan (duplicate, already implemented, validation failure) show generic "No plan created" instead of the agent's explanation.


artem-ivy-ai:
### Staging removed

PR #1192, opened 2026-06-12
fix: type AgentApp prompt with Enter, full-bypass permissions, job-based plan creation

nielsbosma:
- Revert Claude prompt to being typed into the PTY (not a positional CLI arg) so arbitrary content survives; send a separate Enter after a delay so it submits.
- ClaudePty FullAuto now emits --dangerously-skip-permissions instead of --permission-mode dontAsk, which did not bypass command execution.
- AgentPrompt.md + CreatePlanDialog: interactive agent does the suggest/research work first, then creates plans by starting a CreatePlan job (tendril job start CreatePlan) rather than running tendril plan create directly.
- Remove now-unused InitialPrompt plumbing.
- Includes DraftMarkdown annotation popover frontend tweaks.

### DIFF
--- a/src/Ivy.Tendril.Agents.Test/Claude/ClaudePtyTests.cs
+++ b/src/Ivy.Tendril.Agents.Test/Claude/ClaudePtyTests.cs
@@ -155,35 +155,6 @@ public class ClaudePtyTests
     }
 
     [Fact]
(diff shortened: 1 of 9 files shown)


artem-ivy-ai:
### Staging removed

PR #1194, opened 2026-06-12
fix: CreatePlan prompt phrasing/trim and Agent prompt submission

nielsbosma:
- CreatePlanDialog: BuildAgentPrompt trims the description and phrases projects correctly — 'the project X', 'the projects X or Y', and an Auto case that lets the agent pick the project. Trim the description on the direct Create path too. Tests added.
- AgentApp: retry Enter a few times after the bracketed paste so the prompt actually submits (the byte matches a real user Enter; the issue was timing).
- Includes unrelated jobs-cleared migration (Migration_014) and related JobService/PlanDatabaseService changes.


artem-ivy-ai:
### Staging removed

PR #1267, opened 2026-06-18
Include sanitized config.yaml in bug reports

nielsbosma:
## What

Bug reports now attach a redacted copy of `config.yaml` (as `config.sanitized.yaml`) alongside the existing plan/job logs, for both job and plan reports (UI dialog + `report-bug` CLI).

## Why

Reports are uploaded to a **public GitHub issue**, so the config must never go up verbatim. Secrets are stripped before upload:

- `auth.password`, `auth.hashSecret`
- `llm.apiKey`, `api.apiKey`
- all `codingAgents[].environmentVariables` values (keys kept, values → `[REDACTED]`)

## Notes

- Config is read **from disk**, not the expanded in-memory `Settings`, so `%VAR%` references stay literal instead of expanding to their secret values.
- Parsing failures **fail closed** — the config is omitted rather than risk uploading unredacted content.
- `BugReportFile` gained optional in-memory `Content` so the synthesized file is zipped without a temp file on disk.
- Added `BugReportServiceConfigTests` (4 tests, passing).


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-tendril-staging-docs-1267.sliplane.app](https://ivy-tendril-staging-docs-1267.sliplane.app)

Issue #1266, opened 2026-06-18
Bug report: wrong branch again


nielsbosma:
## Bug Report

| Field | Value |
|-------|-------|
| **Report ID** | `3218c774-6edc-4bd7-8cfc-90e1997a5a6b` |
| **Agent** | claude |
| **Tendril Version** | 1.0.51 |
| **OS Version** | Unix 6.17.0.1018 |

## Description

wrong branch again

## Attachment

[Download bug report ZIP](https://sttendrilproduction.blob.core.windows.net/bug-reports/3218c774-6edc-4bd7-8cfc-90e1997a5a6b.zip)

nielsbosma:
Fixed in #1274 (merged to `development`).

Same root cause as #1218: with no `baseBranch` configured, Tendril assumed the default branch was `main` instead of detecting it, so plans on a non-`main` repo targeted the wrong branch.

**Fix:** `GitHelper.ResolveDefaultBranch` now detects the real default branch (local `origin/HEAD` → `git ls-remote --symref` → `"main"` last-resort), and all base-branch resolution sites (worktree creation, CreatePr, sync/preflight, repo-add) use it. A regression test reproduces the exact failure — a clone with `origin/HEAD` unset now resolves the correct branch via the remote instead of falling back to `main`.

Issue #1218, opened 2026-06-15
Bug report: tendril pushes to the wrong branch


nielsbosma:
## Bug Report

| Field | Value |
|-------|-------|
| **Report ID** | `774f6118-010b-45b3-b669-5820e0da2279` |
| **Agent** | claude |
| **Tendril Version** | 1.0.51 |
| **OS Version** | Unix 6.17.0.1018 |

## Description

tendril pushes to the wrong branch

## Attachment

[Download bug report ZIP](https://sttendrilproduction.blob.core.windows.net/bug-reports/774f6118-010b-45b3-b669-5820e0da2279.zip)

nielsbosma:
Fixed in #1274 (merged to `development`).

**Root cause:** when a repo had no `baseBranch` configured, Tendril hardcoded the literal `"main"` as the base branch in five places. On any repo whose real default branch isn't `main` (e.g. `development`/`master`), worktrees were branched from `origin/main` and PRs were opened against the wrong base.

**Fix:** added `GitHelper.ResolveDefaultBranch`, which detects the repo's actual default branch (local `git symbolic-ref refs/remotes/origin/HEAD` → `git ls-remote --symref` → `"main"` only as a last resort), and routed every former `"main"` fallback through it. The detected branch is also persisted into the project config when a repo is added.

Verified with unit tests including live checks against repos whose defaults are `development`, `main`, and `master`.

Issue #1196, opened 2026-06-14
Project `baseBranch` (config.yaml) is not propagated into plan firmware → CreatePr ignores it and targets the repo default branch


rencoivy:
## Summary

A project repo's configured **`baseBranch`** is **not** carried into the plan's firmware (`plan.yaml`). As a result the **CreatePr** promptware — which is explicitly designed to read `baseBranch` from the firmware header — finds nothing and falls back to the repo's **default branch**, so the PR is opened against the wrong base. A plan thus merges to a different branch than the project config declares (and, in our case, than the branch the service was built from).

This is silent: with the Create-PR dialog also not showing the target branch (see #1195), a plan can merge to the wrong branch with zero visibility.

## Environment
- Tendril fork build, project configured with a non-default `baseBranch`.
- Repo default branch: `development`. Project's configured `baseBranch`: `feat/review-pdf-report-tab`.

## Repro
1. Configure a project repo with a `baseBranch` that is **not** the repo's default branch.
2. Run a plan to `ReadyForReview` and Create PR.
3. Observe the PR is opened against the **repo default branch**, not the configured `baseBranch`.

## Evidence (verified in source + live state)

**1. `config.yaml` has `baseBranch` set correctly** (the project repo entry):
```jsonc
{
  "path": "<repo path>",
  "prRule": "default",
  "baseBranch": "feat/review-pdf-report-tab",
  "syncStrategy": "fetch"
}
```

**2. The CreatePr promptware is designed to honor it** — `Promptwares/CreatePr/Program.md`, "Base branch" (~lines 83-87):
> **Base branch:**
> 1. Read plan.yaml and get the project name
> 2. For each repo, check the firmware header for `baseBranch` configuration
> 3. If configured, use that value
> 4. Otherwise, auto-detect via: `gh repo view <owner/repo> --json defaultBranchRef -q .defaultBranchRef.name`

And the command template (~line 77): `gh pr create [...] --base <default-branch> --head <branch> [...]`.

**3. But the plan's firmware (`plan.yaml`) carries NO `baseBranch`** — the `repos:` block is a bare flat list (repo path, then PR URL + commit SHAs), with the per-repo `baseBranch` dropped:
```yaml
repos:
- <repo path>
- https://github.com/<owner>/<repo>/pull/6
- <commit sha>
- <commit sha>
...
```

[...]

PR #1274, opened 2026-06-20
Detect repo default branch instead of assuming "main"

nielsbosma:
## Problem

When a repo had no `baseBranch` configured, Tendril hardcoded the literal `"main"` in five places, so repos whose real default branch differs (e.g. `development`) had worktrees branched from `origin/main` and PRs opened against the wrong base. Fixes the "pushes to the wrong branch" reports (#1218, #1266; also resolves the no-config gap behind #1196).

## Change

- **`GitHelper.ResolveDefaultBranch`** — layered, git-only default-branch detection: local `git symbolic-ref refs/remotes/origin/HEAD` → `git ls-remote --symref` → `"main"` only as last resort. Works for both local clones and not-yet-cloned remote URLs; successful detections are memoized.
- **Detect-and-save at repo-add time** — `ProjectRepoPickerView` and `ProjectAddRepoCommand` now persist the detected default branch into `RepoRef.BaseBranch` when the user doesn't specify one, so project config is explicit going forward.
- **Runtime fallbacks** — replaced all five hardcoded `"main"` assumptions (`JobLauncher` ×2, `JobStartCommand`, `UsePreflightCheck`) with the resolver, covering repos already in config with a null `baseBranch`.
- **De-dup** — `PlanArtifactSyncer.DetectBaseBranch` now delegates to the shared resolver.

## Tests

- `GitHelperTests`: empty-path/no-remote fallback, clone with/without `origin/HEAD` (reproduces the bug → resolves `development` via `ls-remote`), and live-remote checks against Ivy-Framework (`development`), Tendril-Test-Runner (`main`), SeoTools-for-Excel-Connectors (`master`).

Created with the yolo-push skill.

### DIFF
--- a/src/Ivy.Tendril.Test/Helpers/GitHelperTests.cs
+++ b/src/Ivy.Tendril.Test/Helpers/GitHelperTests.cs
@@ -193,6 +193,20 @@ public class GitHelperTests : IDisposable
         Assert.Equal("development", GitHelper.ResolveDefaultBranch(clone));
     }
 
+    [Theory]
+    [InlineData("https://github.com/Ivy-Interactive/Ivy-Framework", "development")]
+    [InlineData("https://github.com/Ivy-Interactive/Tendril-Test-Runner", "main")]
+    [InlineData("https://github.com/nielsbosma/SeoTools-for-Excel-Connectors", "master")]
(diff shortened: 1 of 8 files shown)


artem-ivy-ai:
### Staging removed

PR #1282, opened 2026-06-22
Fix #1257: new drafts not refreshing + harden flaky timing tests

### LINKED ISSUE #1257: New drafts don't appear on Drafts page until app restart (PlanWatcher not refreshing, Windows)
## Summary
A newly created plan **does not appear on the Drafts page until the app is fully restarted**. Reported by a user on Windows 11: after the plan was created, "getting to read the plan itself required a full restart of the app so the plan appeared on the Drafts page."

## Environment
- OS: **Windows 11**

## Expected
Newly created drafts appear on the Drafts page automatically, without restarting the app.

## Notes / pointers
- `src/Ivy.Tendril/Apps/Drafts/DraftsApp.cs` subscribes to `PlanWatcherService` (auto-refresh) and lists plans with status `Draft`/`Blocked`.
- `PlanWatcherService` (`src/Ivy.Tendril/Services/Plans/PlanReaderService.cs` ~L44-73) watches the top-level Plans dir with a 500ms debounce and a 30s poll fallback.
- Since the watcher exists but didn't fire, this is likely a real bug — suspect Windows `FileSystemWatcher` reliability for **newly created** plan directories, and/or the 30s poll not surfacing the new `Draft`. Please repro specifically on Windows.

## Acceptance criteria
- Creating a plan causes it to show up on the Drafts page within the debounce/poll window, no restart required, on Windows.

_Source: direct user feedback._

nielsbosma:
## #1257 — New drafts don't appear on the Drafts page until restart

A newly created plan folder fires the top-level `FileSystemWatcher` while still empty; the later writes *inside* the folder (`plan.yaml`, first revision) don't re-trigger it (watcher is `DirectoryName`-only, no subdirectories). So a draft created outside the GUI's job pipeline could take up to the 30s poll to appear.

**Fix:** after a `Created`/`Renamed`/`Error` event, `PlanWatcherService` schedules a short burst of staggered full re-scans (default +1s/+3s/+8s) so the completed folder is picked up within seconds. The 30s poll remains as the long-stop. Delays are injectable for deterministic tests.

Covered by new `PlanWatcherServiceTests` (a self-heal unit test + a DB-sync integration test that proves an out-of-band plan reaches the DB without a restart).

## Test-suite hardening (deterministic under parallel load)

Audited every timing primitive in the test project and fixed the genuinely flaky ones:

- **`JobService.RunBlockedJobCheck()`** (new internal) — lets the blocked-job dependency test drive the check directly instead of `await Task.Delay(70s)`.
- Fixed sleeps → poll-until-condition (new sync `RetryHelper.WaitUntil`): notify tests flush background writes; inbox race/stagger tests poll observable state; the stagger test asserts the deterministic cumulative-span floor instead of jitter-prone per-gap timing; the sub-resolution 10ms timestamp sleep uses a clock-advance wait.
- `GitServiceDirtyStateTests.AheadOfOrigin` asserts `reason.Commits` (what the code populates) rather than `reason.Files`.
- `WorktreeValidationTests` uses the in-memory-settings `ConfigService` ctor to avoid a shared default `TENDRIL_HOME` race.

Verified: 5 consecutive fully-green full-suite runs (1461 passed, 1 skipped, 0 failed). Worst single test dropped from ~70s to instant.

### DIFF
--- a/src/Ivy.Tendril.Test/Helpers/RetryHelper.cs
+++ b/src/Ivy.Tendril.Test/Helpers/RetryHelper.cs
@@ -2,6 +2,32 @@ namespace Ivy.Tendril.Test.Helpers;
 
 public static class RetryHelper
 {
+    /// <summary>
+    ///     Synchronously polls <paramref name="condition" /> until it returns true or the timeout
+    ///     elapses. Use this instead of a fixed <c>Thread.Sleep</c> before an assertion on async /
+    ///     background / timer / file-watcher work — it removes load-sensitive flakiness by waiting
+    ///     for the observable result rather than guessing how long the work takes.
+    /// </summary>
+    /// <returns>True if the condition became true within the timeout; false otherwise.</returns>
+    public static bool WaitUntil(
+        Func<bool> condition,
+        TimeSpan? timeout = null,
+        TimeSpan? pollInterval = null)
+    {
+        var limit = timeout ?? TimeSpan.FromSeconds(10);
+        var interval = pollInterval ?? TimeSpan.FromMilliseconds(25);
+        var deadline = Environment.TickCount64 + (long)limit.TotalMilliseconds;
+
+        while (Environment.TickCount64 < deadline)
+        {
+            if (condition())
+                return true;
+            Thread.Sleep(interval);
+        }
+
+        return condition();
+    }
+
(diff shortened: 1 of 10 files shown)


artem-ivy-ai:
### Staging removed

PR #1287, opened 2026-06-22
Send git commit SHA as commitId in bug reports

nielsbosma:
## Summary
Tags bug reports with the source revision the client was built from, so a report can be traced back to exact code. Pairs with the backend's optional `commitId` field on `report-bug`.

- **Embed commit at build time** (`Ivy.Tendril.csproj`): a `SetCommitId` MSBuild target emits `[AssemblyMetadata("CommitId", …)]`. Uses CI-passed `-p:CommitId` when present, otherwise falls back to `git rev-parse HEAD` so local/dev builds are tagged too. Hooks `BeforeTargets="GetAssemblyAttributes"` and is tolerant of a missing git (no attribute → field omitted).
- **Read & send** (`BugReportService.cs`): `GetCommitId()` reads the embedded attribute; `UploadAsync` adds the optional `commitId` form field only when non-empty.
- **CI** (`publish-tendril.yml`): pass `-p:CommitId=${{ github.sha }}` to the `dotnet pack` and `dotnet publish` steps.

## Verification
- Local build → embeds current `git rev-parse HEAD`.
- `-p:CommitId=<sha>` override wins over git.
- Build still succeeds with no git (attribute omitted; report uploads without `commitId`).

Dev builds are tagged with their HEAD commit (last commit, not uncommitted working-tree state).

### DIFF
--- a/.github/workflows/publish-tendril.yml
+++ b/.github/workflows/publish-tendril.yml
@@ -123,7 +123,8 @@ jobs:
           dotnet pack src/Ivy.Tendril/Ivy.Tendril.csproj \
             --configuration Release \
             --output ./nupkg \
-            /p:Version=${{ needs.version.outputs.version }}
+            /p:Version=${{ needs.version.outputs.version }} \
+            /p:CommitId=${{ github.sha }}
 
       - name: NuGet login
         uses: NuGet/login@v1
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

Issue #1303, opened 2026-06-22
Bug report: fails to create plan


nielsbosma:
## Bug Report

| Field | Value |
|-------|-------|
| **Report ID** | `478ca4dd-5383-42d2-b523-eb3218d48453` |
| **Agent** | claude |
| **Tendril Version** | 1.0.53 |
| **OS Version** | Unix 26.5.1 |
| **Commit** | `d0c8bc49c849fa6d42fab12e37957612d7845f2c` |

## Description

fails to create plan

## Attachment

The bug report ZIP is stored privately in Azure Blob Storage. Team members can
download it with the Azure CLI (requires `az login` with access to the storage account):

```sh
az storage blob download --account-name sttendrilproduction --container-name bug-reports --name 478ca4dd-5383-42d2-b523-eb3218d48453.zip --file 478ca4dd-5383-42d2-b523-eb3218d48453.zip --auth-mode login
```

Blob path: `https://sttendrilproduction.blob.core.windows.net/bug-reports/478ca4dd-5383-42d2-b523-eb3218d48453.zip`

Issue #1297, opened 2026-06-22
tendril plan create command fails to match project argument


dcrjodle:
## Summary
The `tendril plan create` command frequently fails with an error indicating it cannot match the project name argument.

## Steps to Reproduce
```
tendril plan create "Fix Pagination Density Not Applied to Items" Ivy-Framework \
  --plans-dir "/Users/rorychatt/.tendril/Plans" \
  --level Feature \
  --initial-prompt "Fix that density is not properly applied in the Pagination widget - all three density sizes (Small, Medium, Large) render identically" \
  --execution-profile balanced
```

## Expected vs Actual Behavior
**Expected:** Command should successfully create a plan for the "Ivy-Framework" project.

**Actual:** Command exits with code 1 and error message:
```
Error: Could not match 'Ivy-Framework' with an argument.
```

## Reported by
Mikael Rinne

---

[Slack thread](https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1782128846628689?thread_ts=1782128846.628689&cid=C0AS51ABBB6)

PR #1316, opened 2026-06-22
Fix tendril plan create dash-value parsing (#1297/#1303)

nielsbosma:
## What
`tendril plan create` (and other CLI commands) failed when a free-form option value began with a dash — e.g. a cleaned task description opening with a markdown bullet (`- ...`) or a flag-like word (`--watch ...`). Spectre.Console.Cli reads any token starting with `-` as an option name, so the space-separated form mis-parsed and the command failed ("fails to create plan", #1303). #1297 was a related, already-fixed symptom (project as positional).

## Changes
- **Enforce `--option=value` (equals) form** for every `tendril` CLI option across all promptware `Program.md` (both the bundled `src/Ivy.Tendril/Promptwares` tree and `src/Ivy.Tendril.TeamIvyConfig/Promptwares`), `Prompts/{AgentPrompt,Plans}.md`, plus an authoring rule in both `Promptwares/AGENTS.md`. `gh`/`git`/`dotnet` commands and boolean flags are left in space form.
- **Regression tests** (`PlanCliCommandTests`) drive the real Spectre parser end-to-end: hyphenated project name, and leading-dash `--initial-prompt` in equals form.
- `TestPlanConfigService` gains an optional `projectName` parameter.

## Testing
Full `Ivy.Tendril.Test` suite: 1504 passed, 1 skipped, 0 failed.


artem-ivy-ai:
### Staging removed

PR #1317, opened 2026-06-22
[00560] Job Debug Sheet Hide Empty Permission Denials

dcrjodle:
# Summary

## Changes

Added a null-guard to the Permission Denials field builder in JobDebugSheet. When the denials string is empty, the builder now returns `null` instead of rendering an empty CodeBlock, eliminating unnecessary UI chrome for jobs without permission issues.

## API Changes

None.

## Files Modified

- **src/Ivy.Tendril/Apps/Views/Sheets/JobDebugSheet.cs** — Updated PermissionDenials builder to return null when string is empty

---

## Commits

- 4ddb0798 Hide empty permission denials in Job Debug Sheet


artem-ivy-ai:
### Staging removed

Issue #1291, opened 2026-06-22
Bug report: plan.yaml is not being created when tendril invokes tendril…


nielsbosma:
## Bug Report

| Field | Value |
|-------|-------|
| **Report ID** | `280c50dd-cb80-45c5-9a46-a841d38d4bb5` |
| **Agent** | claude |
| **Tendril Version** | 1.0.53 |
| **OS Version** | Unix 26.5.1 |
            | **Commit** | `4a19ab9986eeea6aaccf76611d13aaa91eb36a26` |

## Description

plan.yaml is not being created when tendril invokes tendril plan create

## Attachment

The bug report ZIP is stored privately in Azure Blob Storage. Team members can
download it with the Azure CLI (requires `az login` with access to the storage account):

```sh
az storage blob download --account-name sttendrilproduction --container-name bug-reports --name 280c50dd-cb80-45c5-9a46-a841d38d4bb5.zip --file 280c50dd-cb80-45c5-9a46-a841d38d4bb5.zip --auth-mode login
```

Blob path: `https://sttendrilproduction.blob.core.windows.net/bug-reports/280c50dd-cb80-45c5-9a46-a841d38d4bb5.zip`

PR #1330, opened 2026-06-22
Add SQLite busy_timeout via shared connection factory (#1291)

nielsbosma:
## Context

Issue #1291: `tendril plan create` failed with "database is locked" when the daemon was running. The Tendril daemon holds a long-lived SQLite connection open while CLI commands (`doctor`, `database`, `run`) open their own connections to the same file. Connections enabled WAL but never set `PRAGMA busy_timeout`, so a writer contending for the lock — specifically during lock escalation inside an explicit write transaction, which Microsoft.Data.Sqlite's ADO-layer retry does **not** cover — surfaced "database is locked" immediately instead of waiting.

(The original CLI-side `plan create` symptom is already resolved on `development`, since the CLI no longer opens the DB. This hardens the remaining daemon/CLI contention paths.)

## Change

- **New `SqliteConnectionFactory`** — centralizes on-disk connection creation, applying `PRAGMA busy_timeout=5000; journal_mode=WAL; foreign_keys=ON` consistently.
- Routed all four connection sites through it: `PlanDatabaseService`, `DatabaseCheck`, `DatabaseCommands`, `RunCommand` (the last previously set no PRAGMAs at all).
- `PlanDatabaseService` keeps its open-raw → integrity-check → recover-if-corrupt ordering; only the post-recovery pragma pass changed (now adds `busy_timeout`).
- Removes duplicated connection setup and two now-unused `using` directives.

## Tests

- New `SqliteConnectionFactoryTests`: asserts the applied pragmas (`busy_timeout=5000`, `journal_mode=wal`) and a fast concurrent-writer serialization smoke test.
- Full unit suite: **1506 passed, 1 skipped (pre-existing), 0 failed**. Build clean (0 warnings); `dotnet format` clean.

### DIFF
--- a/src/Ivy.Tendril.Test/SqliteConnectionFactoryTests.cs
+++ b/src/Ivy.Tendril.Test/SqliteConnectionFactoryTests.cs
@@ -1,0 +1,95 @@
+using Ivy.Tendril.Database;
+using Microsoft.Data.Sqlite;
+
+namespace Ivy.Tendril.Test;
+
+public class SqliteConnectionFactoryTests : IDisposable
+{
+    private readonly TempDirectoryFixture _tempDir = new();
+    private readonly string _dbPath;
+
+    public SqliteConnectionFactoryTests()
+    {
+        _dbPath = Path.Combine(_tempDir.Path, $"factory-test-{Guid.NewGuid():N}.db");
+    }
+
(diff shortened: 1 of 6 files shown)


artem-ivy-ai:
### Staging removed

PR #1341, opened 2026-06-23
Fix tunnel Deactivate requiring two clicks; optimistic UI

nielsbosma:
## Problem
Pressing **Deactivate** on the Tunnel settings required two clicks to take effect.

`Deactivate()` was synchronous and blocked the Ivy dispatcher thread with `supervisorTask.Wait()`. Meanwhile the background tunnel supervisor's `TunnelDisconnected` callback marshalled its `state.Set(...)` updates back to that same blocked thread, so the first click's UI update was never flushed to the client. The second click worked only because the supervisor was already torn down.

## Fix
- `ICloudflaredService`: `void Deactivate()` → `Task DeactivateAsync()`.
- `CloudflaredService`: `await supervisorTask.WaitAsync(...)` instead of a blocking `.Wait()`, so the dispatcher thread is never held. Added `if (ct.IsCancellationRequested) break;` after `WaitForExitAsync` so an intentional stop no longer logs a spurious "exited unexpectedly", re-fires `TunnelDisconnected`, or triggers a restart.
- `TunnelSetupView`: async Deactivate handler that flips state **optimistically** (instant switch to the Activate view) before awaiting the background teardown — mirroring how Activate sets `isLoading` first.

Builds clean (0 warnings, 0 errors).

### DIFF
--- a/src/Ivy.Tendril/Apps/Settings/TunnelSetupView.cs
+++ b/src/Ivy.Tendril/Apps/Settings/TunnelSetupView.cs
@@ -58,7 +58,7 @@ public class TunnelSetupView : ViewBase
 
         if (isLoading.Value)
         {
-            form |= Callout.Info("Starting tunnel and waiting for it to become routable. This typically takes 15-30 seconds.", "Tunnel Starting...");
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #1344, opened 2026-06-23
Tunnel: connecting state, wallpaper QR, and Open in Browser

nielsbosma:
## What

Three related tunnel UX fixes, built on an explicit tunnel status model.

### 1. Show loading while the tunnel is starting
`IsConnected` previously went `true` the moment cloudflared *printed* its URL (in `StartAsync`), **before** the health check confirmed the tunnel was routable. So after a restart the settings view showed "Tunnel Active" while the tunnel wasn't actually reachable yet.

Introduced `TunnelStatus { Disabled, Connecting, Connected }`:
- `CloudflaredService` tracks status and raises a single `StatusChanged` event (replacing `TunnelConnected`/`TunnelDisconnected`). `Connected` is set **only after** `WaitForTunnelHealthyAsync` succeeds.
- `TunnelSetupView` renders by status: `Connecting` → loading spinner, `Connected` → URL/QR/Deactivate, `Disabled` → Activate.

### 2. Tunnel QR on the wallpaper
`WallpaperApp` shows the tunnel QR in a top-right floating panel, but **only when `Status == Connected`** (fully routable) — never during `Connecting`.

### 3. "Open in Browser" button
Added next to "Copy URL" in the tunnel settings, using `client.OpenUrl(...)`.

## Notes
Compiles clean (full build succeeds). No unit tests exist for `CloudflaredService` (only an HTTP-level E2E test, unaffected). Runtime state transitions are best verified in the running app.

### DIFF
--- a/src/Ivy.Tendril/Apps/Settings/TunnelSetupView.cs
+++ b/src/Ivy.Tendril/Apps/Settings/TunnelSetupView.cs
@@ -13,38 +13,25 @@ public class TunnelSetupView : ViewBase
         var tunnelService = UseService<ICloudflaredService>();
         var copyToClipboard = UseClipboard();
 
-        var isLoading = UseState(false);
         var error = UseState<string?>(null);
+        var status = UseState(tunnelService.Status);
         var tunnelUrl = UseState<string?>(tunnelService.TunnelUrl);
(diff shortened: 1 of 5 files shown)


artem-ivy-ai:
### Staging removed

PR #1345, opened 2026-06-23
Tunnel: kill orphaned cloudflared, fix reactivate race and wallpaper menu

nielsbosma:
### 1. Orphaned cloudflared processes (main one)
`Stop()`/`Dispose()` only kill cloudflared when they actually run. On Windows a child isn't killed when its parent dies, so any non-graceful Tendril exit (crash, console close, IDE/terminal force-kill, memory watchdog) left cloudflared orphaned — they were piling up across sessions/days.

Added `ChildProcessTracker`: a Windows **Job Object** with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, created once per process. Every cloudflared process is assigned to it in `TunnelSession.StartAsync`, so the OS terminates them whenever Tendril dies **for any reason**. Best-effort; no-op on non-Windows (there we still rely on `Stop()`/`Dispose()`).

### 2. Deactivate → reactivate race
The old and new supervisor shared the `_currentSession` field, so a superseded supervisor's `finally` could dispose the **newly started** session — killing a fresh tunnel before it registered with Cloudflare's edge (`DNS_PROBE_FINISHED_NXDOMAIN`, while the UI reported "ready"). Each supervisor generation now owns a local session and only clears the shared field when it still references its own (`ReferenceEquals`).

### 3. Wallpaper tunnel dropdown did nothing
`WithDropDown`'s default handler routes a click by matching the selected value against each `MenuItem`'s `Tag`/`Label`; the items had no `Tag`, so nothing matched. Added explicit tags to **Copy to Clipboard** / **Open in Browser** / **Deactivate**.

## Notes
Builds clean. No unit tests for `CloudflaredService` (only an HTTP-level E2E test, unaffected). Best verified manually: deactivate→reactivate a few times (URLs should work), and hard-kill Tendril while a tunnel is active (cloudflared should die with it).

### DIFF
--- a/src/Ivy.Tendril/Apps/WallpaperApp.cs
+++ b/src/Ivy.Tendril/Apps/WallpaperApp.cs
@@ -14,9 +14,11 @@ public class WallpaperApp : ViewBase
 {
     public override object Build()
     {
+        var client = UseService<IClientProvider>();
         var versionService = UseService<IVersionCheckService>();
         var planDbService = UseService<IPlanDatabaseService>();
         var tunnelService = UseService<ICloudflaredService>();
+        var copyToClipboard = UseClipboard();
         var versionInfo = UseState<VersionInfo?>(null);
         var dismissedVersion = UseState<string?>(null);
         var tunnelStatus = UseState(tunnelService.Status);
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #1346, opened 2026-06-23
Tunnel: only report Connected when actually routable

nielsbosma:
## Problem
The tunnel was reported **Connected** before it was actually routable:
- After 60s of failed probes the health check logged "proceeding anyway" and **returned successfully**, so the supervisor flipped to Connected anyway → "says ready but doesn't work".
- It only treated `502` as not-ready, and 60s was too short for cloudflare DNS propagation. An early probe against a not-yet-published hostname returns `NXDOMAIN`, which the OS resolver **caches negatively** and keeps returning past the negative-TTL window even after the record goes live — so it only worked "if you wait even longer".

## Rework of `WaitForTunnelHealthyAsync`
- **Wait 3s before the first probe** so a premature `NXDOMAIN` doesn't poison the resolver's negative cache.
- **Poll up to 180s** (was 60s), with per-attempt logging.
- **Recognize all not-ready codes** — `502`, `504`, `530` (Argo/Tunnel error 1033). Any other completed response = the request reached the origin through the tunnel = routable.
- **Throw on real timeout instead of returning.** The supervisor then tears the session down and retries with a fresh hostname (no poisoned negative cache) instead of leaving a dead "Connected" tunnel. The UI stays on the Connecting spinner until the URL genuinely works.

## Notes
Builds clean. Best verified manually: activate a tunnel and confirm it stays on the spinner until the URL actually loads (no premature "Active"), and tolerates slow DNS.

### DIFF
--- a/src/Ivy.Tendril/Services/Tunnel/CloudflaredService.cs
+++ b/src/Ivy.Tendril/Services/Tunnel/CloudflaredService.cs
@@ -124,24 +124,41 @@ public sealed class CloudflaredService : ICloudflaredService, IStartable, IDispo
         _supervisorTask = Task.Run(() => SupervisorLoopAsync(_cts.Token));
     }
 
+    private static readonly TimeSpan HealthCheckTimeout = TimeSpan.FromSeconds(180);
+    private static readonly TimeSpan HealthCheckInterval = TimeSpan.FromSeconds(2);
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #1364, opened 2026-06-25
Fix duplicate Default option in coding agent profile dropdowns

nielsbosma:
## What

The "Profile Models" dropdowns (Deep / Balanced / Quick) in the Coding Agent settings listed both **Default** and **OpenCode Default** when OpenCode was selected.

## Why

`CodingAgentSetupView` prepends a hardcoded `("Default", "default")` option to the agent's catalog models. OpenCode's catalog uniquely includes a model with `Id = "default"` / `DisplayName = "OpenCode Default"`, so both entries (sharing the same `default` value) showed up. Other agents have no `default`-id model, so they were unaffected.

## Fix

Filter out catalog models whose `Id` is `"default"`, since the hardcoded "Default" option already represents that exact value. Only "Default" now appears for OpenCode; other agents are unchanged.


artem-ivy-ai:
### Staging removed

PR #1365, opened 2026-06-25
Expose tendril CLI to AgentApp PTY agent via shim

nielsbosma:
## What

The **Agent** app (`AgentApp`) launches a coding agent in a PTY via `Context.UsePty`. When running a plan with Claude in dev — where there is no installed `tendril` binary on PATH — the agent could not invoke `tendril ...` commands.

## Why

Promptware/job launches (`PromptwareRunner`, `JobLauncher`) already make `tendril` available by calling `AgentProcessHelper.EnsureTendrilOnPath` (creates a shim that proxies to `dotnet exec Ivy.Tendril.dll`) and setting `TENDRIL_HOME/CONFIG/PLANS` on their `ProcessStartInfo`. AgentApp's PTY path was the only launcher that skipped this — its `GetEnvironment` only returned the user's per-agent env vars.

## Fix

- Extracted the shim-dir creation out of `EnsureTendrilOnPath` into a reusable `EnsureTendrilShimDir()` (behavior of the existing `ProcessStartInfo` path is unchanged).
- Added `ApplyTendrilEnvironment(IDictionary<string,string>, IConfigService)` that sets `TENDRIL_HOME/CONFIG/PLANS` and prepends the shim dir to `PATH` — the dictionary-based equivalent of what promptwares do.
- `AgentApp.GetEnvironment` now applies it to the dictionary passed to `UsePty`. `UsePty` inherits the full parent environment and merges this dictionary on top, so the agent gets the `tendril` shim on PATH plus the active config/plans.

No Ivy-Framework changes were needed. Verified with `dotnet build` (0 warnings, 0 errors).

### DIFF
--- a/src/Ivy.Tendril/Apps/Agent/AgentApp.cs
+++ b/src/Ivy.Tendril/Apps/Agent/AgentApp.cs
@@ -1,6 +1,7 @@
 using Ivy.Hooks.Pty;
 using Ivy.Tendril.Agents.Abstractions;
 using Ivy.Tendril.Agents.Helpers;
+using Ivy.Tendril.Helpers;
 using Ivy.Tendril.Services;
 using Ivy.Widgets.Xterm;
 using Xterm = Ivy.Widgets.Xterm;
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #1370, opened 2026-06-25
Add Dialogs entry to Settings > Debug submenu

nielsbosma:
## What

`DialogsApp` (the DirtyRepoDialog test harness) was unreachable from the UI.

## Why

It's marked `[App(isVisible: false)]` so it never shows in the app list, and the **Settings → Debug** submenu in `TendrilAppShell` was hardcoded to only list **Onboarding**.

## Fix

Added a **Dialogs** menu item that navigates to `DialogsApp`, alongside Onboarding (DEBUG builds only, like the rest of that submenu).


artem-ivy-ai:
### Staging removed

PR #1376, opened 2026-06-25
Polish DirtyRepoDialog and add extensible dialog test harness

nielsbosma:
## What

Polish for `DirtyRepoDialog` plus a refactor of its debug harness so more dialog tests can be added.

## Dialog rendering
- Entire content renders as **Markdown**: bold repo name, repo path and file paths as inline code, reasons as a bulleted list.
- Commit shas render as inline code while the subject stays plain text (`` `a1b2c3d` Fix null ref ``).
- Proper pluralization computed in the dialog — `1 uncommitted change` / `N uncommitted changes` (no lazy `(s)`). Doing it in the dialog (not per-`DirtyReasonDetail.Message`) keeps real and test data rendering identically.
- Dropped the `---` rule between repos.

## Bug fix
The dialog double-stripped `UncommittedChanges` paths — `GitService` already strips the `git status --porcelain` status prefix, so **live usage rendered `/Program.cs`** instead of `src/Program.cs`. The test harness hid it by injecting a fake ` M ` prefix. Removed the extra strip and made the harness emit realistic unprefixed paths, so both render correct repo-relative paths.

## Labels
Renamed proceed labels at the call sites to **Create/Execute Without Syncing**.

## Harness refactor
- Extracted `DirtyRepoDialogDebugView` (all the scenario builders + preview).
- `DialogsApp` is now a thin shell that wraps each dialog test in an `Expandable` — add a new `*DebugView` + one list entry to grow it.

## Tracking fix
`src/Ivy.Tendril/Apps/Debug/` was being matched by the `[Dd]ebug/` build-output rule in `.gitignore`, so `DialogsApp.cs` was never tracked. Added a negation to keep the source folder tracked (`DialogsApp.cs` is added in this PR).

Verified with `dotnet build` (0 errors).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ x86/
 [Oo]ut/
 [Bb]uild[Ll]og.*
 
+# Source folder literally named "Debug" — not build output; keep it tracked.
+!src/Ivy.Tendril/Apps/Debug/
+
 ## NuGet
 *.nupkg
 **/[Pp]ackages/*
--- a/src/Ivy.Tendril/Apps/Debug/DialogsApp.cs
+++ b/src/Ivy.Tendril/Apps/Debug/DialogsApp.cs
@@ -1,0 +1,27 @@
+namespace Ivy.Tendril.Apps.Debug;
+
+// Debug-only harness for visually reviewing how the various Tendril dialogs render across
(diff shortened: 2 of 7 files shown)


artem-ivy-ai:
### Staging removed

PR #1385, opened 2026-06-25
Polish DetailsTab: capitalize Profile, hide empty Related Plans / Depends On

nielsbosma:
## What

Two small fixes to the plan **Details** tab.

### Profile capitalization
Execution profiles are stored lowercase (`balanced`, `deep`, `quick`). The Details row now title-cases the value for display (`balanced` → `Balanced`). Empty profiles stay empty so `RemoveEmpty()` still drops the row.

### Hide empty Related Plans / Depends On
These rows were showing even when empty. Root cause: Ivy-Framework's `ValidationHelper.IsEmptyContent` (used by `Details.RemoveEmpty()`) only treated `null`/blank-string/`false` as empty — an empty `List<Link>` counted as non-empty. The framework fix (handle empty `IEnumerable`) is in the Ivy-Framework repo and will reach Tendril on the next Ivy package bump.

Until then, this PR resolves `RelatedPlans`/`DependsOn` to `null` (instead of an empty list) when there are no entries, which the current `RemoveEmpty()` already drops. Once the fixed Ivy package is referenced this becomes redundant and can be simplified back to plain `ParsePlanLinks(...)`.

Verified with `dotnet build` (0 errors, 0 warnings).


artem-ivy-ai:
### Staging removed

PR #1395, opened 2026-06-25
Assign current user as assignee in CreatePr

nielsbosma:
The CreatePr promptware now passes `--assignee @me` to `gh pr create`, so PRs it opens are assigned to the current (gh-authenticated) user. Self-assignment is documented as non-fatal so a repo that rejects the assignee won't break PR creation. Applied to both the built-in and TeamIvyConfig copies of the promptware.


artem-ivy-ai:
### Staging removed

PR #1462, opened 2026-06-29
Persist dismissed update version across restarts

nielsbosma:
The "Update Available" notification's **Dismiss** button only updated in-memory state, so the panel reappeared on every restart.

- Add `TendrilSettings.DismissedUpdateVersion` (persisted to `config.yaml` via `SaveSettings()`).
- `WallpaperApp` seeds the dismiss state from config and writes the dismissed version on Dismiss, so it stays dismissed across restarts. A newer release still surfaces because the gate compares against the specific version string.
- Includes misc Copilot provider/test updates present in the working tree.


artem-ivy-ai:
### Staging removed

PR #1466, opened 2026-06-29
Fix Next button disabled after setup agent completes (#1450)

### LINKED ISSUE #1450: Next button greyed out and unclickable after project setup with Codex
## Summary
After completing project setup with Codex, the "Next" button appears greyed out and is unclickable, preventing progression through the setup flow.

## Additional Context
The issue occurs after attempting to configure project wiring through the CLI using commands like:
```
tendril project add-verification
tendril project add-review-action
```

While these commands return success, the configuration does not persist to the saved project config. Verification of the YAML shows that `verifications` and `reviewActions` fields for BSHemsida remain empty, indicating the setup is only partially complete due to a Tendril CLI persistence bug.

A detailed reflection of this issue was recorded at: `/Users/ksg25/.tendril/Promptwares/SetupProject/Memory/cli-persistence-quirks-2026-06-29.md`

The setup step itself shows as "Completed" with token usage displayed (1,245,117 in / 11,317 out), but the Next button remains disabled.

## Related
[Slack thread](https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1782724840483079?thread_ts=1782724840.483079&cid=C0AS51ABBB6)

Reported by Sethcool

---

dcrjodle:
## Problem

During onboarding, after the `SetupProject` agent finishes (the `AgentViewer` shows "Completed" with token counts), the **Next** button stays greyed out and unclickable, blocking the setup flow. Reported on Codex CLI runs. Fixes #1450.

**Root cause:** The Next button is disabled by `session.Running.Value`, which is only cleared in the `finally` of a background task that awaits `handle.Completion`. That completion ultimately awaits `process.WaitForExitAsync` in `PromptwareRunner.PipeOutputAndLog` **with no timeout**. The agent can emit its terminal `result` event (which is what makes the `AgentViewer` render "Completed") while keeping the output pipe open, so `session.Running` never transitions to `false` and the button stays disabled. The "config doesn't persist" symptom is a *consequence* of never reaching the `finally` (which calls `ReloadSettings()`), not a serialization bug — the CLI commands persist correctly.

## Fix

`src/Ivy.Tendril/Apps/Onboarding/ProjectAgentStepView.cs`

1. **Wire `AgentViewer.OnComplete`.** The frontend `AgentViewer` raises `OnComplete` the moment it parses a terminal `result` event from the stream. The handler now clears `session.Running` / `isStepLoading`, so the button enables as soon as the agent reports completion — independent of process exit. The background task's `finally` still runs `ReloadSettings()` whenever the process eventually exits, so config reload is unaffected; both paths clearing the flags is idempotent.

2. **Clear `session.Running` in the error path.** The outer `catch` already reset `isStepLoading` but not `session.Running`; an exception after `Running.Set(true)` could otherwise leave the button permanently disabled.

## Deviations from the original (Codex) plan

- **Did not add a wall-clock timeout to `PipeOutputAndLog`.** The proposed flat 2-minute `CancelAfter` conflated stale-output vs. total-runtime semantics and would kill legitimately long agent runs. The `OnComplete` fix resolves the UI symptom without that risk.
- **Did not add a new `ProjectCommandTests` test.** `AddVerification_AddsToProject` already covers the CLI write → fresh-`ConfigService` reload round-trip.

## Verification

- `dotnet build src/Ivy.Tendril` — succeeded, 0 warnings/errors
- `dotnet build src/Ivy.Tendril.Test` — succeeded
- `dotnet test --filter ProjectCommandTests` — 18/18 passed
- `dotnet format` — clean

### DIFF
--- a/src/Ivy.Tendril/Apps/Onboarding/ProjectAgentStepView.cs
+++ b/src/Ivy.Tendril/Apps/Onboarding/ProjectAgentStepView.cs
@@ -215,6 +215,7 @@ public class ProjectAgentStepView(
                 error.Set($"Failed to set up project: {ex.Message}");
                 isCloning.Set(false);
                 isStepLoading.Set(false);
+                session.Running.Set(false);
             }
         }, setupTrigger != null ? [setupTrigger, EffectTrigger.OnMount()] : [EffectTrigger.OnMount()]);
 
@@ -235,6 +236,21 @@ public class ProjectAgentStepView(
         // that avoided a layout shift when the bordered/padded Box swapped in on first output.
         var showStream = !isCloning.Value && (session.Running.Value || session.HasOutput.Value);
 
+        var viewer = new AgentViewer()
+            .Stream(session.Stream)
+            .AutoScroll(true)
+            .ShowStatusLabel(true)
+            .Width(Size.Full())
+            .Height(Size.Full()) with
+        {
+            OnComplete = _ =>
+            {
+                session.Running.Set(false);
+                isStepLoading.Set(false);
+                return ValueTask.CompletedTask;
+            }
+        };
+
         return Layout.Vertical().Margin(0, 0, 0, 2)
                | (showHeader ? Text.H3("Setting up your project") : null!)
                | Text.Muted(isCloning.Value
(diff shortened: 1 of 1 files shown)


artem-ivy-ai:
### Staging removed

PR #1467, opened 2026-06-29
Remove dead state-reset code from UseTrigger-mounted dialogs

nielsbosma:
Follow-up to the CreatePrDialog cleanup. These dialogs are mounted via `UseTrigger`, which unmounts the dialog view (disposing all its `UseState`) when closed and remounts a fresh instance on the next open. The state resets performed right before closing were therefore dead code — writing to state about to be discarded.

Each close handler now only sets its open-state to false.

**Cleaned:**
- `ImportIssuesDialog` — removed 3 `ResetDialogState()` calls + the now-unused helper
- `UpdatePlanDialog` — 3 reset blocks
- `SuggestChangesDialog` — 3 reset blocks
- `AcceptWithNotesDialog` — 3 `notesText` resets
- `RerunJobDialog` — `feedback` reset in `Close()`
- `ResetToDraftDialog` — `isResetting` reset in onClose

Left untouched: `CreateIssueDialog` (mixed local + parent-owned state) and `AgentTestDialog` (persistently mounted, resets are load-bearing).


artem-ivy-ai:
### Staging removed

PR #1482, opened 2026-06-29
Make stackHash format shell-safe (/ and .lang instead of | and parens)

nielsbosma:
## Problem

Agents setting the stack hash via `tendril project set <name> stackHash "<hash>"` fail on Windows. The agent wraps the command as `cmd /c "… "fe…|be…" …"`; cmd.exe collapses the inner quotes and parses `|` as a pipe → `'be' is not recognized` (exit 255). The format used `|` (segment separator) and `(lang)` (qualifier) — both shell metacharacters.

## Fix

Make the hash contain only shell-safe characters (letters, digits, `+ - _ . : /`) so it survives any shell/agent **unquoted**:
- segment separator `|` → `/`
- language qualifier `(ts)` → `.ts`

```
fe(ts):react+next+tailwind|be(ts):nestjs+prisma|db:postgres
->  fe.ts:react+next+tailwind/be.ts:nestjs+prisma/db:postgres
```

The hash is an **opaque string** everywhere (storage, display, telemetry) — nothing parses it — so this is a spec/value change only, no code logic touched.

### Changes
- **`SetupProject/Program.md`** (shipped + TeamIvyConfig): updated grammar, all reference examples, added a **Shell-safety** note (so the rule isn't reverted), and dropped the now-unnecessary quotes in the instruction (`stackHash <hash>` — the quotes were what cmd mangled).
- **`TeamIvyConfig/config.yaml`**: updated the one stored hash to the new format.

Verified no code splits on `|` / extracts `(lang)`; no remaining old-format artifacts in tracked source (historical Logs left as-is).


artem-ivy-ai:
### Staging removed

Issue #1454, opened 2026-06-29
Remove outdated code now that updater is implemented


dcrjodle:
## Summary

Remove code that is no longer needed now that the updater functionality has been implemented.

## Additional Context

The image shows the Ivy Tendril application with an "Update Available" notification displaying version v1.1.2 with an update command. This suggests the updater system is now in place and working.

---

**Slack thread:** https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1782729409021729?thread_ts=1782729409.021729&cid=C0AS51ABBB6

Reported by @dcrjodle

---

### Attachments

![1782729409-021729-Screenshot 2026-06-29 at 12.36.37.png](https://stivytelemetry.blob.core.windows.net/dumpsterbot/a49c6782-504e-402d-be72-47182a75f7aa/1782729409-021729-Screenshot_2026-06-29_at_12.36.37.png)

Issue #1368, opened 2026-06-25
Consider removing Ivy.Tendril.Updater


dcrjodle:
## Summary
Question raised about whether `Ivy.Tendril.Updater` should be removed from the project.

## Additional Context
The `Ivy.Tendril.Updater` project is currently part of the Ivy.Tendril solution (as shown in the project structure).

## Related Discussion
https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1782375144837939?thread_ts=1782375144.837939&cid=C0AS51ABBB6

Reported by @nielsbosma, @rorychatt

---

### Attachments

![1782375144-837939-image.png](https://stivytelemetry.blob.core.windows.net/dumpsterbot/79c0870b-9c0e-41a5-8bcb-7f03889b61f4/1782375144-837939-image.png)

PR #1485, opened 2026-06-30
feat: one-click Velopack update + Check for updates in Settings

dcrjodle:
## Summary

Consolidates the in-app update experience onto a single **Velopack** self-update backend and adds a **Check for updates** action to the Settings dropdown. This is PR 1 of 2 — behavior/UX only, nothing is deleted here.

The in-app update toast already had an "Update Now" button, but it sat alongside three overlapping update mechanisms (Velopack self-update, an installer re-download fallback, and a never-invoked separate updater GUI). This PR makes the toast drive **one** backend and exposes the same check from Settings.

## Changes

- **`VersionCheckService` / `IVersionCheckService`**
  - `bool CanSelfUpdate` — true only for Velopack installs (`VelopackLocator.Current?.CurrentlyInstalledVersion != null`).
  - `Task StartUpdateAsync(UpdateProgress, CancellationToken)` — the Velopack-only download + `ApplyUpdatesAndRestart` flow, shared by the toast and (the trigger for) Settings.
  - `CheckForUpdatesAsync(bool forceRefresh = false)` — lets "Check for updates" bypass the 15-minute cache.
- **`WallpaperApp` (update toast)** — refactored to call `StartUpdateAsync`; the **Update Now** button is now gated on `CanSelfUpdate` (non-Velopack installs see version info + the copy-paste install command, but no button that can't work); the installer-redownload branch is removed from the UI. Progress / error / dismiss states preserved.
- **`TendrilAppShell`** — new **Check for updates** item in the Settings dropdown; force-refreshes the version check and reports the result via a toast (available / up-to-date / failed).

## Result

The only thing a user does to update is click **Update Now** (Velopack downloads and restarts the app) — or, for non-Velopack installs, run the shown command. The CLI `tendril update` is unchanged.

## Notes

- The CLI fallback (installer re-download for non-Velopack installs) is intentionally kept — acceptable in a terminal context.
- **PR 2 (follow-up)** removes the dead `Ivy.Tendril.Updater` project and its build/embed plumbing.

## Verification

- `dotnet build src/Ivy.Tendril/Ivy.Tendril.csproj` → succeeds, 0 warnings / 0 errors.
- Not yet smoke-tested in a running app (requires a real Velopack install to exercise `CanSelfUpdate`).

Refs #1454, #1368

### DIFF
--- a/src/Ivy.Tendril/AppShell/TendrilAppShell.cs
+++ b/src/Ivy.Tendril/AppShell/TendrilAppShell.cs
@@ -120,6 +120,7 @@ public class TendrilAppShell(AppShellSettings settings) : ViewBase
         var selectedIndex = UseState<int?>();
         var appRepository = UseService<IAppRepository>();
         var client = UseService<IClientProvider>();
+        var versionService = UseService<IVersionCheckService>();
         var currentApp = UseState<AppHost?>();
         var statusService = UseService<ITendrilProcessStatusService>();
         var agentRunner = UseService<IAgentRunner>();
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

Issue #48, opened 2026-04-16
(git flow) Confusing 'Create PR' button when only local files are modified


dcrjodle:
## Summary
The "Create PR" button appears in the UI even when only local files (such as tendril config) are modified, which is confusing because no PR will actually be created in this scenario.

## Additional Context
- Slack thread: https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1776347706921279

## Reported by
Mikael Rinne

nielsbosma:
If there are no commits in the worktrees - then maybe we should disable the "Create PR".

PR #1489, opened 2026-06-30
Fix macOS codesigning JIT entitlements (#48)

rorychatt:
This PR fixes the macOS codesigning issue (#48) where nested runtimes (dotnet/PowerShell) are signed without entitlements, causing them to be terminated with SIGKILL under Hardened Runtime. It generates a temporary entitlements plist and signs all Mach-O executables with it.


artem-ivy-ai:
### Staging removed

Issue #1455, opened 2026-06-29
Bug report: never started, only said "running..." for ever


nielsbosma:
## Bug Report

| Field | Value |
|-------|-------|
| **Report ID** | `97f4f8e3-c206-4913-8092-3fd958cc5818` |
| **Agent** | claude |
| **Tendril Version** | 1.1.1 |
| **OS Version** | Unix 6.17.0.1018 |
| **Commit** | `e8fd691fd220c626fa9ed0d2a65bf23f1fd20eae` |

## Description

never started, only said "running..." for ever

## Attachment

The bug report ZIP is stored privately in Azure Blob Storage. Team members can
download it with the Azure CLI (requires `az login` with access to the storage account):

```sh
az storage blob download --account-name sttendrilproduction --container-name bug-reports --name 97f4f8e3-c206-4913-8092-3fd958cc5818.zip --file 97f4f8e3-c206-4913-8092-3fd958cc5818.zip --auth-mode login
```

Blob path: `https://sttendrilproduction.blob.core.windows.net/bug-reports/97f4f8e3-c206-4913-8092-3fd958cc5818.zip`

PR #1497, opened 2026-07-01
Fix jobs hanging "running..." forever with timeouts never firing (#1455)

nielsbosma:
## Context

Bug report #1455: an ExecutePlan job ran ~97 min producing no output and no commits, showing "running…" the whole time, until manually Stopped — the 30-min job timeout, 35-min hard timeout, and 10-min stale-output timeout all failed to fire.

Root cause: work happens in an **un-monitored pre-launch window** where a hang can't be timed out or Stopped. Three defects in that window are fixed here.

## Changes

**`JobLauncher.StartAgentProcess`** — wrote the whole compiled prompt to stdin *synchronously* before draining stdout/stderr and before the timeout monitor was armed → classic three-pipe deadlock, and no timeout could ever fire. Now drains output first and writes stdin fire-and-forget (reusing `ProcessRunner.WriteStdinAndCloseAsync`) so the monitor arms immediately; non-ignorable write failures are logged rather than silently faulting a detached task.

**`JobMonitor.RunStaleOutputWatchdog`** — never tripped for a job that produced no output (`LastOutputAt` stayed null → skipped forever). Now anchors the stale window to when monitoring begins, so a zero-output hang is killed after `staleOutputTimeout`, without counting slow pre-launch hook time against the window.

**`JobCompletionHandler` hook execution** (`RunHookAction` / `EvaluateHookCondition`) — read stdout to EOF *before* the `WaitForExitOrKill` guard, making the timeout dead code, and read streams sequentially (deadlock risk; the condition path never drained stderr at all). Now reads both streams concurrently before a real bounded wait, so a hung/flooding `before` hook is force-killed at its guard (10s condition / 30s action) instead of wedging the launch.

## Tests

Regression tests for all three paths in `JobServiceTimeoutTests`, `JobLauncherTests`, and `JobServiceHookTests`. Each reproduces a hang against the old code and passes against the fix. Build is clean (`--warnaserror`); all 206 job tests pass.

## Note

Also carries a few small **unrelated** UI tweaks that were already in the working tree: `RerunJobDialog` placeholder text, a width removal in `JobsApp.DataTable`, and a `CliCommand` field in `JobDebugSheet`.


artem-ivy-ai:
### Staging removed

PR #1502, opened 2026-07-01
Fix missing gaps between review action buttons in ReviewApp

nielsbosma:
The horizontal layout holding the review action buttons in the ReviewApp used `.Gap(0)`, so multiple action buttons rendered flush against each other with no spacing.

Changed to `.Gap(2)` to match the sibling control/title bars in the same view (`ContentView.cs`).


artem-ivy-ai:
### Staging removed

PR #1518, opened 2026-07-02
Add optional GitHub username to bug reports

nielsbosma:
Adds an optional "Your GitHub Username" input to the Report Bug dialog and a matching `--github-user` option to the `report-bug` CLI command.

The value is forwarded through `BugReportService.SubmitReportAsync` to the new optional `githubUser` form field on the report-bug endpoint, and is only sent when non-empty (mirroring the existing `commitId` handling). Existing callers and empty submissions behave exactly as before.


artem-ivy-ai:
### Staging removed

PR #1523, opened 2026-07-02
Add first-class promptware failure reporting via `tendril job fail`

nielsbosma:
## What

Gives promptwares a first-class way to report a hard failure (e.g. a worktree that couldn't be created) instead of `exit 1`-and-hope-the-host-guesses.

- **New CLI command** `tendril job fail <job-id> --message "..."` → `PUT /api/jobs/{id}/fail` → `JobService.ReportJobFailure` records it on `JobItem.ReportedFailureReason`.
- **`SetCompletionStatus`** now uses the declared reason when the job completes non-zero, winning over any stale progress message set via `tendril job status` and over the `ExtractFailureReason` output-scraper (which stays as the fallback for promptwares that just `exit 1`).
- **Firmware** injects a "Failure Reporting" instruction into every promptware; **ExecutePlan `Program.md`** wires the call into the worktree-isolation and worktree-creation failure paths (main + TeamIvyConfig copies).
- **Cleanup:** shared discover/serialize/PUT boilerplate extracted into `MasterClient.PutJson`; both `JobStatusCommand` and `JobFailCommand` route through it.

## Why

The old path relied on `JobFailureAnalyzer.ExtractFailureReason` heuristically scraping output, which often surfaced an unrelated line (e.g. a stray `fatal:`) instead of the real cause.

## Tests

- `JobServiceFailureReasonTests`: declared reason wins over a prior progress message and over scraping; falls back to the scraper when no reason is reported; `ReportJobFailure` set/not-found.
- `JobFailCommandTests`: settings validation.
- Updated `IJobService` test stubs for the new interface member.

### DIFF
--- a/src/Ivy.Tendril.TeamIvyConfig/Promptwares/ExecutePlan/Program.md
+++ b/src/Ivy.Tendril.TeamIvyConfig/Promptwares/ExecutePlan/Program.md
@@ -78,6 +78,7 @@ if [ -f .git ] && grep -q "gitdir:" .git; then
     echo "ERROR: Repository at <repo-path> is itself a worktree."
     echo "ExecutePlan cannot create worktrees inside worktrees."
     echo "Check that project repo paths point to main repositories, not worktrees."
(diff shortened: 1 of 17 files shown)


artem-ivy-ai:
### Staging removed

PR #1540, opened 2026-07-02
Make job deletion optimistic (background worktree cleanup)

nielsbosma:
## Optimistic job deletion (background worktree cleanup)

Deleting an `ExecutePlan` job whose plan is non-terminal previously blocked the UI thread on `git worktree remove` and recursive Windows directory deletes (seconds, worst case tens of seconds), leaving the confirm dialog and job row on screen until cleanup finished.

`JobService.ApplyDeletePlanState` now runs cleanup on a background task, so `DeleteJob` returns immediately and the Jobs UI updates optimistically via the existing `JobsStructureChanged -> refreshToken.Refresh()` bridge.

- Cleanup runs **before** `ResetToDraft`, so the plan only becomes re-runnable after its worktree/artifact directories are gone — closing a race where an immediate re-execution would collide with the still-running deletes.
- Cleanup failures are surfaced via the existing `NotificationReady` destructive toast instead of being silently swallowed.
- Test `DeleteJob_ExecutePlan_NonTerminal_ResetsToDraft` polls for the now-asynchronous reset.

## Tooling (unrelated, bundled)

- `DisableGit.ps1`: `RepoPath` is now a required positional parameter (safer than defaulting to the current directory for a git-auth footgun); docs/examples updated.
- `AnalyzeFailed.ps1`: new promptware helper that scans `*.raw.jsonl` logs for failed shell commands.

Verified: `dotnet build` clean; targeted test suites (deletion, worktree cleanup, completion-guard) pass (84 tests).

### DIFF
--- a/src/Ivy.Tendril.Test/JobServiceCompletionGuardTests.cs
+++ b/src/Ivy.Tendril.Test/JobServiceCompletionGuardTests.cs
@@ -97,6 +97,12 @@ public class JobServiceCompletionGuardTests : IDisposable
 
         service.DeleteJob(id);
 
+        // Cleanup + ResetToDraft now run on a background thread (optimistic delete), so the
+        // reset lands after DeleteJob returns rather than synchronously.
+        var deadline = DateTime.UtcNow.AddSeconds(5);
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #1565, opened 2026-07-02
fix(ci): fix syntax error in macOS postinstall script

rorychatt:
This PR fixes the unexpected EOF syntax error in the macOS postinstall script generation by removing the single quote from the comment.


artem-ivy-ai:
### Staging removed

PR #1566, opened 2026-07-02
Add tendril config get/set to CLI and MCP

nielsbosma:
## What

Adds a `tendril config get/set` CLI group and `tendril_get_config` / `tendril_set_config` MCP tools to read/write top-level `config.yaml` settings — closing the gap where the CLI/MCP couldn't do what the Settings UI does.

Keys: `codingAgent`, `jobTimeout`, `staleOutputTimeout`, `gitTimeout`, `maxConcurrentJobs`, `planTemplate`. CLI and MCP share one field switch (`ConfigCommand.ReadField`/`ApplyField`) — no parallel drift.

## Highlights

- **Validated writes:** int bounds mirror `ConfigService.ValidateSettings`; `codingAgent` is validated against the registered agents and stored as the canonical id; empty `codingAgent` rejected. Confirmation reports the value actually stored.
- **Long values:** `set` accepts `--file`/`--stdin` for multiline `planTemplate`; `get` writes raw so it round-trips cleanly.
- **Bug fix (data loss):** `planTemplate` was path-normalized and re-persisted at load, so any save (even an unrelated field, or the UI) corrupted slashes/URLs. It now stays raw in `Settings` and is expanded at point of use in `JobLauncher`; `VariableExpansion.ExpandVariables` gained a `normalizePaths` opt-out.
- MCP host now registers agent infrastructure so `ConfigTools` can validate.
- Docs: new CLI Config page, MCP Configuration tools table, AGENTS/GEMINI references.

Also includes a Create Plan dialog layout tweak and AgentPrompt plan-creation guidance updates.

## Tests

New `ConfigCliCommandTests` + `ConfigToolsTests` (get/set round-trips, slash/URL round-trip, unrelated-set-preserves-template, bounds/empty/unknown rejection, codingAgent validation + canonicalization). Full suite green apart from pre-existing, machine-cache-dependent `VersionCheckServiceTests`.

### DIFF
--- a/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/00_Overview.md
+++ b/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/00_Overview.md
@@ -10,6 +10,7 @@ searchHints:
   - run
   - doctor
   - version
+  - config
 ---
 
 # CLI Overview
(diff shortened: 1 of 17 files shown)


artem-ivy-ai:
### Staging removed

PR #1569, opened 2026-07-03
feat: implement self-update capability and dialog

rorychatt:
This PR implements a beautiful, interactive Update Tendril Dialog that allows users to seamlessly check for updates and update Tendril from the App Shell. It integrates with the existing version check service and triggers appropriate user actions depending on whether self-update is supported for their installation type.


rorychatt:
@claude review this

artem-ivy-ai:
### Staging removed

PR #1612, opened 2026-07-03
[00051] Fix Shared Event Parser Race by Isolating Parsers Per Session

nielsbosma:
# Summary

## Changes

Fixed a race condition where `AgentRunner` kept one shared `IEventParser` instance per agent id and reused
it across concurrent sessions via `Reset()`, corrupting `_hasError` and the accumulated cost/token totals
when two jobs of the same agent type ran simultaneously. Replaced the reset-and-reuse pattern with
create-fresh-per-session: `AgentRunner.GetParser` now returns a brand-new parser instance from a write-once
prototype instead of mutating and returning the shared one.

## API Changes

- `IEventParser.Reset()` removed; `IEventParser.CreateFresh()` added, returning a new independent instance.
- All six parser implementations (`OpenCodeEventParser`, `AntigravityEventParser`, `ClaudeEventParser`,
  `CodexEventParser`, `GeminiEventParser`, `CopilotEventParser`) implement `CreateFresh()` as `new XxxParser()`
  instead of `Reset()`.
- `AgentRunner.GetParser(string agentId)` behavior change: now returns a fresh instance per call instead of
  the same shared, reset instance. `Register` overloads are unchanged.

## Files Modified

- `src/Ivy.Tendril.Agents/Abstractions/IEventParser.cs` — interface change
- `src/Ivy.Tendril.Agents/Runtime/AgentRunner.cs` — `GetParser` now calls `CreateFresh()`
- `src/Ivy.Tendril.Agents/Providers/{OpenCode,Antigravity,Claude,Codex,Gemini,Copilot}/*EventParser.cs` — `Reset()` → `CreateFresh()`
- `src/Ivy.Tendril.Agents.Test/Runtime/AgentRunnerRegistrationTests.cs` — new `GetParser_ReturnsFreshInstancePerCall` regression test
- `src/Ivy.Tendril.Agents.Test/OpenCode/OpenCodeEventParserTests.cs` — new `CreateFresh_ReturnsIndependentInstance` test
- 6 parser test files — removed 7 obsolete `Reset()`-specific tests

## Manual Testing

N/A — internal fix to session-scoped parser lifecycle with no user-facing UI surface. Correctness is covered
by the two new regression tests: `AgentRunnerRegistrationTests.GetParser_ReturnsFreshInstancePerCall` proves
two `GetParser` calls for the same agent id return distinct instances, and
`OpenCodeEventParserTests.CreateFresh_ReturnsIndependentInstance` proves a fresh instance doesn't inherit
another instance's accumulated cost/token state.

---
Commits:
- 5ce4fb7e Fix shared event parser race by isolating parsers per session
- c199e9d9 Add parser-per-session regression tests, remove obsolete Reset() tests

---

### DIFF
--- a/src/Ivy.Tendril.Agents.Test/Antigravity/AntigravityEventParserTests.cs
+++ b/src/Ivy.Tendril.Agents.Test/Antigravity/AntigravityEventParserTests.cs
@@ -67,19 +67,6 @@ public class AntigravityEventParserTests
     }
 
     [Fact]
-    public void Reset_ResetsInitializedState()
-    {
-        _parser.ParseLine("Line 1");
-        _parser.Reset();
-
-        // After reset, the next line should trigger SessionInit again
-        var events = _parser.ParseLine("Line 2");
-        Assert.Equal(2, events.Count);
-        Assert.IsType<SessionInitEvent>(events[0]);
-        Assert.IsType<TextEvent>(events[1]);
-    }
-
-    [Fact]
     public void BuildResult_WithExistingResultEvent_UpdatesExitCode()
     {
         var events = new List<AgentEvent>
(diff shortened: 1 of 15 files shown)


artem-ivy-ai:
### Staging removed

PR #1629, opened 2026-07-06
Job Debug agent focus prompt + recommendation row badges

nielsbosma:
## Job Debug sheet
- **Debug with <agent>** now opens a dialog first, with an optional *"Anything particular you want to look for?"* textbox. The note is folded into the agent prompt (`In particular, focus on: ...`); leaving it blank keeps the previous behavior. New `DebugWithAgentDialog`. (DEBUG builds only.)
- Both dialogs (Report Bug + Debug with Agent) now use the `UseTrigger` pattern with hooks at the top of `Build()`, satisfying the Ivy rules-of-hooks analyzer.
- Unified the details view and the copy/debug text onto a single `JobDebugData` record, so field values are computed once. The confirm path snapshots details while the job is live, so it always navigates even if the job is evicted after the dialog opens.

## Recommendations
- Show the **Project** + **Impact** badges on each sidebar row (matching the Drafts/Icebox/Review sidebars) and removed them from the detail-header title.

### DIFF
--- a/src/Ivy.Tendril/Apps/Jobs/Dialogs/DebugWithAgentDialog.cs
+++ b/src/Ivy.Tendril/Apps/Jobs/Dialogs/DebugWithAgentDialog.cs
@@ -1,0 +1,32 @@
+namespace Ivy.Tendril.Apps.Jobs.Dialogs;
+
+public class DebugWithAgentDialog(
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

PR #1630, opened 2026-07-06
Extract Review tabs into dedicated views

nielsbosma:
Refactors `Review/ContentView.cs` so each tab is its own view with clear separation of concerns.

- **New** `Review/Tabs/PlanTabView.cs` — read-only plan markdown tab.
- **New** `Review/Tabs/RecommendationsTabView.cs` — Implement button + recommendation row list; `RecommendationRowView` moves in alongside its only consumer.
- **New** `Review/ReviewActionsBarView.cs` — project review-actions bar (not a tab, so directly under `Review/`).
- `ContentView.BuildContent` now composes these views while retaining the tab assembly, navigation, tested static helpers, and overlay sheets.

Shared `Views/Tabs/*` views are untouched (Drafts still depends on them). No behavior change — build clean, `ContentViewTests` (14) green. ContentView.cs shrinks ~860 → ~797 lines.


artem-ivy-ai:
### Staging removed

PR #1643, opened 2026-07-06
[00056] Switch Open Review and Drafts Tabs to the Clicked Plan

nielsbosma:
# Summary

## Changes

Fixed a bug where clicking a plan id in Jobs to navigate to Review or Drafts only focused the tab
(keeping the previously-selected plan) instead of switching to the clicked plan, when the target tab
was already open. Added a `RefreshExistingTab` route action that replaces the existing tab's `AppHost`
with one built from the new navigation args and forces a remount (via a fresh `RefreshToken`) whenever
the incoming args differ from the tab's current args. The fix lives in the shared `AppShellRouter` /
`TendrilAppShell` code, so it applies to both Review and Drafts (and any other tabbed app) without
per-app changes.

## API Changes

- `AppShellRouter.RouteAction` (internal enum) — new member `RefreshExistingTab`, inserted between
  `SwitchToExistingTab` and `CreateNewTab`.
- `TendrilAppShell` — new private local function `HandleRefreshExistingTab(NavigateArgs, int, string,
  bool)`, wired into the `OpenApp` switch statement.

Both are internal implementation details of the Tendril app shell; no public API surface changed.

## Files Modified

- `src/Ivy.Tendril/AppShell/AppShellRouter.cs` — duplicate-tab branch of `RouteForTabs` now compares
  serialized args JSON and returns `RefreshExistingTab` when they differ (and the incoming args are
  non-null), otherwise `SwitchToExistingTab` as before.
- `src/Ivy.Tendril/AppShell/TendrilAppShell.cs` — added the `RefreshExistingTab` case and
  `HandleRefreshExistingTab` handler, which swaps the tab's `AppHost` and `RefreshToken`, updates the
  selected index/title, and redirects on `HistoryOp.Push`.
- `src/Ivy.Tendril.Test/AppShell/AppShellRouterTests.cs` — added 3 tests covering different-args
  (→ `RefreshExistingTab`), same-args (→ `SwitchToExistingTab`), and null-args (→
  `SwitchToExistingTab`) cases.

## Manual Testing

1. Run the Tendril app.
2. Open the **Review** tab (or **Drafts**) and select a plan — note which plan is shown.
3. Without closing that tab, go to **Jobs** and click a *different* plan id whose target state routes
   to the already-open tab (Review for `Review`/`Failed` plans, Drafts for `Draft`/`Blocked` plans).
4. Expected: the open tab switches to show the newly clicked plan (previously it stayed on the old
   plan and only the tab gained focus). Repeat for the other tab (Drafts vs Review) to confirm both
   work.
5. As a regression check: click the sidebar entry for Review/Drafts directly (no plan id) while a tab
   is already open — it should just focus the tab and keep showing whatever plan was previously
   selected (in-tab state preserved), confirming the null-args path still takes the focus-only route.

## Fix: Extract shared AppHost-build-and-redirect helper for tab handlers

Per reviewer feedback, `HandleRefreshExistingTab` and `HandleCreateNewTab` in `TendrilAppShell.cs` each
built an `AppHost` via `navigateArgs.ToAppHost(args.ConnectionId)`, called `SetAppTitle`, and redirected
on `HistoryOp.Push` — the same three steps, duplicated across both handlers. Extracted a new private
local function `BuildAppHostAndRedirect(NavigateArgs, string appId, string tabId, bool replaceHistory)`
that performs all three steps and returns the built `AppHost`; both handlers now call it instead of
repeating the logic inline.

### Files Modified

- `src/Ivy.Tendril/AppShell/TendrilAppShell.cs` — added `BuildAppHostAndRedirect` helper; updated
  `HandleRefreshExistingTab` and `HandleCreateNewTab` to use it.

## Fix: Unify tab-replace idiom between SetItem and RemoveAt+Insert

Per reviewer feedback, `OnTabRefresh` in `TendrilAppShell.cs` replaced a tab in place using
`tabs.Value.RemoveAt(tabIndex).Insert(tabIndex, ...)`, while `HandleRefreshExistingTab` did the
conceptually identical "swap tab in place" operation with `tabs.Value.SetItem(tabIndex, ...)`.
Changed `OnTabRefresh` to use `SetItem` as well, so both call sites use the same idiom for the same
operation.

### Files Modified

- `src/Ivy.Tendril/AppShell/TendrilAppShell.cs` — `OnTabRefresh` now uses `tabs.Value.SetItem(...)`
  instead of `RemoveAt(...).Insert(...)`.

---
## Commits

- 3680b19c [00056] Unify tab-replace idiom on ImmutableArray.SetItem in OnTabRefresh
- c3971450 code-review: preserve original state-then-redirect ordering in shared helper
- 67be8381 [00056] Extract shared AppHost-build-and-redirect helper for tab handlers
- 9d4587ef Add RefreshExistingTab routing tests (Plan 00056)
- ef97f200 Refresh existing tab args on navigate (Plan 00056)

---

### DIFF
--- a/src/Ivy.Tendril.Test/AppShell/AppShellRouterTests.cs
+++ b/src/Ivy.Tendril.Test/AppShell/AppShellRouterTests.cs
@@ -1,6 +1,7 @@
 using System.Collections.Immutable;
 using Ivy.Core.Apps;
 using Ivy.Tendril.AppShell;
+using Ivy.Tendril.Apps.Review;
 using static Ivy.Tendril.AppShell.TendrilAppShell;
 
 namespace Ivy.Tendril.Test.AppShell;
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #1650, opened 2026-07-10
[00053] Fix Jobs Hanging Forever in Starting State

nielsbosma:
# Summary

## Changes

Jobs could get stranded `Running` forever in the "Starting…" UI state: the launch path marked a job
`Running` (and its plan `Executing`) before `JobMonitor.Start()` ever armed the stale-output/hard
timeout watchdogs, so any unhandled exception in that window left the job un-rescuable. Fixed by
wrapping `JobLauncher.LaunchJob`'s body in try/catch (failing the job cleanly and releasing its
concurrency slot on any exception), moving the plan-to-Executing transition to after the agent
process actually starts (so most launch failures never touch plan state), adding a global
`RunStuckJobCheck` reaper in `JobService` that scans all `Running` jobs on the existing 60s timer
independent of whether their monitor ever armed, and hardening `JobMonitor`'s
`ObjectDisposedException` handler to complete a still-`Running` job instead of returning silently.
A follow-up fix (found during this execution's own code-review pass) also kills the orphaned agent
process if a launch failure occurs after the process already started.

## API Changes

- `JobService.RunStuckJobCheck()` — new `internal` method; global stuck-job reaper, exposed for
  deterministic testing. Called automatically from the existing 60s blocked-job-check timer.
- `JobLauncher`: new private `TransitionPlanToExecuting` and `HandleLaunchFailure` methods (internal
  implementation detail, not part of any public surface).

## Files Modified

- `src/Ivy.Tendril/Services/Jobs/JobLauncher.cs` — guarded launch path, moved plan-to-Executing
  transition, added launch-failure handling with process cleanup.
- `src/Ivy.Tendril/Services/Jobs/JobService.cs` — added `RunStuckJobCheck` reaper, wired into the
  existing periodic timer.
- `src/Ivy.Tendril/Services/Jobs/JobMonitor.cs` — hardened `ObjectDisposedException` recovery to
  complete a still-Running job (with process kill) instead of returning silently.
- `src/Ivy.Tendril.Test/Services/JobLauncherTests.cs`, `src/Ivy.Tendril.Test/JobServiceTimeoutTests.cs`
  — new regression tests for both safety nets.

## Manual Testing

1. In the Jobs view, trigger a launch failure (e.g. temporarily make a "before" hook throw, or
   corrupt a plan's `plan.yaml` so it can't be read) and start an `ExecutePlan` job against it.
2. Observe the job immediately shows `Failed` with a `"Launch failed: ..."` message instead of
   hanging in "Starting…" — and the plan itself is not stuck in the `Executing` state.
3. Start another job of the same type right after — it should launch normally (the concurrency slot
   was not leaked).
4. (Harder to reproduce manually) A job artificially left `Running` with no output for longer than
   `staleOutputTimeout` + 2 minutes will be automatically reaped by the background reaper within the
   next 60-second timer tick, transitioning to `Timeout` with a "No output for N minutes" message.

---
## Commits

- 1dbc7150 [00053] Guard job launch path and add stuck-job reaper
- 296b2c27 [00053] Add tests for launch-failure handling and stuck-job reaper
- f76eb6cb code-review: kill orphaned process when launch fails after process start

---

### DIFF
--- a/src/Ivy.Tendril.Test/JobServiceTimeoutTests.cs
+++ b/src/Ivy.Tendril.Test/JobServiceTimeoutTests.cs
@@ -371,6 +371,50 @@ codingAgent: claude
         service.CompleteJob(id, 0);
     }
 
+    // Regression for #plan-00053: a job stranded Running with no monitor ever armed (e.g. because
+    // the launch hung before JobMonitor.Start ran) previously had no path to completion — none of
+    // the per-job watchdogs exist to rescue it. RunStuckJobCheck is the global safety net that scans
+    // all Running jobs using StartedAt as the baseline, independent of whether a monitor ever started.
+    [Fact]
+    public void RunStuckJobCheck_StaleRunningJobWithNoMonitor_ReapsAsTimeoutWithStaleReason()
+    {
+        var service = CreateService(TimeSpan.FromMinutes(30), TimeSpan.FromMinutes(10));
+
+        var id = service.CreateTestJob(new ExecutePlanArgs(_tempDir.Path));
+        var job = service.GetJob(id);
+        Assert.NotNull(job);
+
+        // Simulate a job stuck in "Starting…": launched long ago, never emitted any output.
+        job.StartedAt = DateTime.UtcNow.AddMinutes(-20);
+        job.LastOutputAt = null;
+
+        service.RunStuckJobCheck();
+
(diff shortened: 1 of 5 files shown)


artem-ivy-ai:
### Staging removed

PR #1652, opened 2026-07-10
[00037] Fix plan.yaml Read-During-Write Race Crashing Job Launch

nielsbosma:
# Summary

## Changes

Fixed a read-during-write race where `PlanCommandHelpers.ReadPlan` could read `plan.yaml` mid-truncate
during a background rewrite by `PlanReaderService`, deserialize an empty file to `null`, and throw an
`InvalidOperationException` that crashed job launch. `ReadPlan` now acquires the same `PlanFileLock`
every writer already holds (serializing reads against writes) plus a bounded retry as defense-in-depth.
Also hardened `UsePreflightCheck`'s background check so an exception there can no longer become an
unobserved task exception or leave the Execute button's spinner stuck forever. The plan's third fix item
(making `JobLauncher.LaunchJob` fail cleanly and release its job-slot semaphore on an unhandled
exception) was found already implemented and tested by a prior, overlapping plan (00041, merged
2026-07-06) — verified present, not re-implemented.

## API Changes

None — both changes are internal robustness fixes with no public API surface change.

## Files Modified

- `src/Ivy.Tendril/Helpers/PlanCommandHelpers.cs` — `ReadPlan` now acquires `PlanFileLock` and retries (up to 5x, 50ms) on empty/null content before throwing.
- `src/Ivy.Tendril/Hooks/UsePreflightCheck.cs` — wrapped the background check in try/catch/finally; logs failures via `CrashLog` and always resets `isChecking`.
- `src/Ivy.Tendril.Test/PlanCommandHelpersReadPlanRaceTests.cs` (new) — regression test reproducing the race via a concurrent writer loop using the same lock/truncate pattern as production code.

## Manual Testing

N/A — internal robustness fix with no directly observable UI change. The regression test
(`PlanCommandHelpersReadPlanRaceTests`) exercises the fix end-to-end: it was confirmed to fail with the
exact reported `InvalidOperationException: Failed to deserialize plan.yaml` on the pre-fix code, and to
pass after the fix, by temporarily reverting the fix commit during this execution and re-running the
test both ways.

---
## Commits

- 3f48a87e Prevent unobserved exception and stuck spinner in UsePreflightCheck (Plan 00037)
- 2351d43a Fix plan.yaml read-during-write race in ReadPlan (Plan 00037)

---

### DIFF
--- a/src/Ivy.Tendril.Test/PlanCommandHelpersReadPlanRaceTests.cs
+++ b/src/Ivy.Tendril.Test/PlanCommandHelpersReadPlanRaceTests.cs
@@ -1,0 +1,57 @@
+using Ivy.Tendril.Helpers;
+
+namespace Ivy.Tendril.Test;
+
+public class PlanCommandHelpersReadPlanRaceTests : IDisposable
+{
+    private readonly string _planFolder;
+
+    public PlanCommandHelpersReadPlanRaceTests()
+    {
+        _planFolder = Path.Combine(Path.GetTempPath(), $"tendril-readplan-race-{Guid.NewGuid()}");
+        Directory.CreateDirectory(_planFolder);
+        File.WriteAllText(Path.Combine(_planFolder, "plan.yaml"), PlanYaml("Test Plan"));
+    }
+
+    public void Dispose()
+    {
+        if (Directory.Exists(_planFolder))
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

PR #1667, opened 2026-07-10
Consolidate all job logs under <TendrilHome>/Jobs/

nielsbosma:
Job artifacts were scattered across three roots — the plan folder's `Logs/`, the promptware's `Logs/` (which resolves into the **source tree** in a dev checkout), and `<TendrilHome>/Jobs/` for the eventwire. The compiled prompt was never stored durably. Debugging one job meant visiting three directories.

## The new layout

Every job writes four files, flat, into `<TendrilHome>/Jobs/`, sharing one stem:

```
Jobs/
  .counter
  00458-00044-ExecutePlan.md              <- Job Log
  00458-00044-ExecutePlan.prompt.md       <- Job Prompt   (new artifact)
  00458-00044-ExecutePlan.raw.jsonl       <- Job Raw Log
  00458-00044-ExecutePlan.eventwire.jsonl <- Job Eventwire Log
  00459-CreatePlan.md                     <- no plan id: not plan-connected
```

The stem derives from `JobItem.PlanFile` alone and is **fixed for a job's lifetime**, so no artifact is ever renamed. A `CreatePlan` job never gains a plan-id segment: `JobCompletionHandler` rewrites its `PlanFile` at completion, which would otherwise scatter its artifacts across two names. Its log records the plan it produced via a `**PlanId:**` header line — the only link back to that plan, and what lets a plan's bug report attach the job that authored it.

## Logs now survive a crash

The raw and eventwire streams are **appended as output arrives** rather than persisted at completion, so a job killed mid-run keeps everything it produced. Verified by `SIGKILL` against a live agent: the completion write never ran, yet the raw and eventwire files were intact.

Job logs are never deleted. Deleting or purging a job removes it from the UI and the database, not its forensic record.

## Breaking: `plan add-log` moves to the job namespace

It always wrote to a job, never to a plan.

| Before | After |
|---|---|
| `tendril plan add-log <plan-id> <action>` | `tendril job add-log <job-id> <action>` |
| MCP `tendril_plan_add_log` | `tendril_job_add_log` |
| `POST /api/plans/{planId}/logs` | `POST /api/jobs/{jobId}/logs` |

It appends an `## Agent Log` section that the completion write preserves, keyed off an HTML-comment marker an agent cannot accidentally emit.

`Promptwares/*/Logs/` and `<Plan>/Logs/` are gone, with no migration — old logs stay on disk but become invisible to the UI and bug reports. Pre-upgrade **eventwire** files *are* still read (flat `{jobId}.eventwire.jsonl` fallback), so no completed job loses its output pane.

## Bugs found and fixed along the way

- **`VersionCheckService.CanSelfUpdate`** read `VelopackLocator.Current`, which **throws** (not returns null) when `VelopackApp.Build()` hasn't run. Inside the `try` it silently degraded to "couldn't check for updates"; `UpdateTendrilDialog` and `WallpaperApp` read it *outside* any try.
- **`JobLogWriter` cleared `job.CliCommand`** before `PersistJob` persisted it, so the `Migration_013` column was always `NULL` and the Debug sheet's "Arguments" row was always empty after completion. The writer no longer mutates the job.
- **`JobService.CreateTestJob` never set `job.Project`**, so `RunHooks` resolved hooks via `GetProject("")` and silently ran **none**.
- **Reading a live job's log threw a sharing violation** — `File.ReadAllLines` requests `FileShare.Read`, which denies the appender its write handle.
- **CLI and REST disagreed on job-id grammar** (`job add-log 458` failed, `POST /api/jobs/458/logs` worked). Both now share `Helpers.JobId`.
- **`JobLogPaths` resolution created directories as a side effect**, and yielded a relative `Jobs` path when `TendrilHome` was unset.
- **`AnalyzeFailed.ps1`** crashed on the single-log case (`Set-StrictMode` + scalar `.Count`), mis-reported the file count, and flagged any command whose output merely *mentioned* `Exit code 1`.

## Tests

**1754 + 950 passing, zero failures**, stable across repeated runs. All 22 pre-existing failures in `Ivy.Tendril.Test` and all 3 in `Ivy.Tendril.Agents.Test` are fixed — most were tests calling `StartJob` on a `JobService` with no `IConfigService`, where the launch can never succeed. One test was pruned as obsolete (`AntigravityModelCatalogTests.GetStaticModels_ReturnsEmpty`, asserting an empty catalog that has since been deliberately populated).

## Known, deliberately not fixed

- `JobLauncher.FailJobAndReleaseSlot` claims completion without calling `HandleCompletion`, so a job that fails at launch runs **no `after` hooks and writes no Job Log** — even though its `before` hooks already ran. Fixing this reaches plan-state reverts, notifications and dependent-job retries; it deserves its own change.
- `BugReportService` reads every `*.md` in the never-pruned `Jobs/` directory when building a plan report. Grows unbounded.
- `PromptwareRunner` pumps stdout and stderr on two threads into a single Rx `Subject` without synchronization.


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-tendril-staging-docs-1667.sliplane.app](https://ivy-tendril-staging-docs-1667.sliplane.app)

Issue #1340, opened 2026-06-23
Bug report: job created from wrong project, and when in review no chang…


nielsbosma:
## Bug Report

| Field | Value |
|-------|-------|
| **Report ID** | `cbef1057-8b55-4665-bda4-8ea8c41b07ad` |
| **Agent** | claude |
| **Tendril Version** | 1.0.51 |
| **OS Version** | Unix 6.17.0.1018 |

## Description

job created from wrong project, and when in review no changes could be seen in changes tab

## Attachment

The bug report ZIP is stored privately in Azure Blob Storage. Team members can
download it with the Azure CLI (requires `az login` with access to the storage account):

```sh
az storage blob download --account-name sttendrilproduction --container-name bug-reports --name cbef1057-8b55-4665-bda4-8ea8c41b07ad.zip --file cbef1057-8b55-4665-bda4-8ea8c41b07ad.zip --auth-mode login
```

Blob path: `https://sttendrilproduction.blob.core.windows.net/bug-reports/cbef1057-8b55-4665-bda4-8ea8c41b07ad.zip`

dcrjodle:
user created plan accidently in wrong project, which caused this bug

PR #1686, opened 2026-07-14
fix: sync commits from every plan worktree, not just plan.Repos matches

### LINKED ISSUE #1670: (review): missing git&changes in review app

dcrjodle:
Fixes #1670.

**Symptom:** After confirming a plan, the review app showed no diff/Changes tab even though the agent had committed.

**Root cause:** `PlanArtifactSyncer.IterateWorktrees` filtered worktree directories under `<PlanFolder>/Worktrees/` by folder name against `plan.Repos` — a snapshot taken at plan-creation time. When the worktree folder name the agent actually worked in didn't match (project pointing at a different/renamed repo, or config changed after plan creation), the worktree was silently skipped, `plan.Commits` stayed empty, and `PlanContentHelpers.GetAllChangesData` returned null before its #1340 "wrong project" fallback could run — so the Changes tab vanished instead of showing the warning banner.

**Fix:** Remove the name filter so all plan worktrees are scanned for commits — consistent with `GitTabDataBuilder`, which already scans all worktrees. Commit attribution remains safe via branch scoping (`git log origin/<base>..tendril/<planFolderName>`) and per-hash dedup. This makes the existing #1340 "wrong project" banner reachable in the mismatch scenario.

**Note on #281:** the removed filter was originally added to suppress phantom commits from read-only build-dependency worktrees. Verified: a dep worktree cut fresh from `origin/<base>` yields zero commits from the branch-scoped log, so no phantom commits reappear in the normal case. If an agent ever commits into a dep worktree, that commit now surfaces via the #1340 unlisted-worktree banner rather than being silently suppressed.

**Testing:** New regression test `SyncPlanArtifacts_WorktreeFolderNameMismatchesPlanRepos_StillSyncsCommits` using a real git repo + worktree fixture — verified it fails against pre-fix code and passes post-fix, plus a matching-name control test. Full suite: 1757 passed, 0 failures.

### DIFF
--- a/src/Ivy.Tendril.Test/PlanArtifactSyncerTests.cs
+++ b/src/Ivy.Tendril.Test/PlanArtifactSyncerTests.cs
@@ -1,0 +1,114 @@
+using Ivy.Tendril.Helpers;
+using Ivy.Tendril.Services.Plans;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Ivy.Tendril.Test;
+
+public class PlanArtifactSyncerTests : IDisposable
+{
+    private readonly string _tempDir = Path.Combine(Path.GetTempPath(), "PlanArtifactSyncerTests_" + Guid.NewGuid().ToString("N")[..8]);
+
+    public PlanArtifactSyncerTests()
+    {
+        Directory.CreateDirectory(_tempDir);
+    }
+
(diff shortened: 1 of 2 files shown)


artem-ivy-ai:
### Staging removed

PR #1693, opened 2026-07-14
feat: highlight selected sidebar list item with filled background

dcrjodle:
## What

Selected items in the master-detail sidebar lists (Review, Drafts, Icebox, Trash, Recommendations) and the selected section in SettingsApp now get a persistent highlighted background until another item is selected.

## Why

- Sidebar list rows rendered as Ghost buttons, so there was no visual indication of which plan/file/recommendation was currently selected — e.g. in ReviewApp you couldn't tell which draft app the content pane belonged to.
- The Settings sidebar used the framework `SidebarMenu`, whose active-item highlight is derived from the URL app id — section tags never match it, so the selected section was never highlighted either.

## How

- `SidebarListRow.Build` (the shared helper behind the sidebars) takes a new `isSelected` flag: selected rows render as a `Secondary` button (filled `bg-secondary`), unselected rows stay `Ghost`. `bg-secondary` is exactly what the framework's own sidebar uses for its active item, so the look is consistent.
- Each sidebar passes its selection state: Review/Drafts/Icebox compare by plan `FolderName`, Trash by file path, Recommendations by record equality.
- SettingsApp's sidebar is rebuilt with the same `SidebarListRow` pattern (new icon-capable overload), keeping the "Configuration" header, the section icons, and the "Open config.yaml" action row.

## Notes for reviewers

- Ivy 1.3.8's `ListItem`/`SidebarMenu` widgets have no app-controllable selected state, hence the button-variant approach on the Tendril side. A `Selected` prop in Ivy-Framework would be the longer-term home for this.
- The highlight follows the existing auto-reselect behavior (first-item auto-select, selection surviving a plan leaving the filtered list) since the sidebars rebuild on any selection change.

### DIFF
--- a/src/Ivy.Tendril/Apps/Drafts/SidebarView.cs
+++ b/src/Ivy.Tendril/Apps/Drafts/SidebarView.cs
@@ -78,7 +78,8 @@ public class SidebarView(
             }
             badges |= new Badge(plan.Level).Color(config.GetLevelColor(plan.Level) ?? Colors.Gray).Small();
 
-            return SidebarListRow.Build($"#{plan.Id} {plan.Title}", badges, () => selectedPlanState.Set(clickablePlan));
+            return SidebarListRow.Build($"#{plan.Id} {plan.Title}", badges, () => selectedPlanState.Set(clickablePlan),
+                plan.FolderName == selectedPlanState.Value?.FolderName);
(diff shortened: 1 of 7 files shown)


artem-ivy-ai:
### Staging removed

PR #1770, opened 2026-07-25
[00061] Redirect Stale .NET Tool Invocations to the Installed Tendril CLI

### LINKED ISSUE #1745: Tendril CLI invokes outdated version when updating from .NET tool to installer
## Summary

When Tendril is installed as a .NET tool and then updated to the installer version, there are cases where Tendril commands are invoked outside of context and users get the outdated version of the Tendril CLI.

## Workaround

Uninstalling the old .NET tool resolves the issue.

## Additional Context

This appears to be a version conflict issue between the .NET tool installation and the newer installer-based installation.

---

Reported by @rorychatt

https://ivy-interactive.slack.com/archives/C0AS51ABBB6/p1784904399314359?thread_ts=1784904399.314359&cid=C0AS51ABBB6

rorychatt:
Fixes #1745

# Summary: Redirect Stale .NET Tool Invocations to the Installed Tendril CLI

Implements the fix for [Issue #1745](https://github.com/Ivy-Interactive/Ivy-Tendril/issues/1745):
when Tendril was installed both via `dotnet tool install -g Ivy.Tendril` and later via the
standalone installer, PATH resolution could favor the stale .NET tool copy, silently running an
outdated CLI with no diagnostic to explain why.

## Changes

1. **`src/Ivy.Tendril/Helpers/TendrilInstallHelper.cs`** (new) — detects whether the running
   process is the legacy .NET tool (`IsLegacyDotnetToolProcess`), reads its installed version
   (`GetLegacyToolVersion`), checks whether it's installed at all (`IsLegacyToolInstalled`),
   locates an installer-managed CLI binary across Windows/macOS/Linux install locations while
   rejecting symlinks that resolve back into the legacy tool tree (`FindInstalledCli`), and scans
   `PATH` for a named executable (`ResolveOnPath`). All path builders honor an injectable
   `UserProfileOverride` for test isolation.
2. **`src/Ivy.Tendril/Program.cs`** — added `TryRedirectLegacyToolInvocation`, called right after
   `AugmentPath` and before `EnsureCliSymlink`. When the running process is the legacy tool and an
   installer-managed CLI is found, it prints a warning, forwards the original arguments to the
   installed binary with a recursion guard (`TENDRIL_NO_LEGACY_REDIRECT=1`), and returns its exit
   code. Any failure falls through to normal execution so the redirect can never break the CLI.
3. **`src/Ivy.Tendril/Helpers/PathHelper.cs`** — reordered `AugmentPath`'s `commonDirs` so
   `~/.local/bin` (where the installer's CLI symlink lives) is added before `~/.dotnet/tools`,
   so the installer's binary wins PATH resolution when both are present.
4. **`src/Ivy.Tendril/Commands/DoctorChecks/InstallationCheck.cs`** (new) — a new `tendril doctor`
   check, registered first (before `EnvironmentCheck`), that reports the running version,
   executable path, PATH resolution status, and flags a conflicting legacy .NET tool install with
   the exact uninstall command to run.
5. **`README.md`** — added a note under `### .NET Tool` explaining that the .NET tool and
   installer are separate installs and that switching from tool to installer requires
   `dotnet tool uninstall --global Ivy.Tendril`.

## Tests

Added 8 new test cases (5 in `TendrilInstallHelperTests.cs`, 1 in `PathHelperTests.cs`, 2 in
`DoctorChecksTests.cs`) covering legacy-tool detection, version resolution, installed-CLI lookup
(including the symlink-rejection case), PATH scanning, PATH ordering, and the new doctor check's
error/ok states. Machine-dependent assertions include defensive skip guards for directories or
fixed OS-standard install paths that may already exist on the running machine.

## Verification

| Verification | Result |
|---|---|
| DotnetFormat | Pass |
| DotnetBuild | Pass |
| DotnetTest | Pass (1797 passed, 1 pre-existing unrelated skip) |
| VitePlusTest | Skipped (no frontend files changed) |
| CheckResult | Pass |

The top-level `Ivy.Tendril.slnx` could not restore in this isolated worktree because it
conditionally references a sibling `Ivy-Framework` checkout not present here; format/build were
run against the two owning `.csproj` files directly instead, per each verification's own documented
fallback.

## Commits

- `5644b68d` Redirect stale .NET tool invocations to the installed Tendril CLI
- `81269f31` Add doctor check for conflicting legacy .NET tool install
- `c0e0c96d` Add tests for legacy .NET tool detection and PATH ordering
- `b303212f` Document the .NET tool / installer conflict in README

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ dotnet tool install -g Ivy.Tendril
 
 > **Tip:** PowerShell 7, Git and gh CLI need to be present on your machine if you install using the `dotnet tool` command.
 
+> **Note:** The `dotnet tool` install and the Quick Install above are separate installs that can both end up on `PATH`. If you move from the .NET tool to the installer, run `dotnet tool uninstall --global Ivy.Tendril` so `tendril` on `PATH` cannot resolve to the older, frozen tool copy. Run `tendril doctor` to check for a conflicting install.
+
 ### Run
 
 ```bash
--- a/src/Ivy.Tendril.Test/DoctorChecksTests.cs
+++ b/src/Ivy.Tendril.Test/DoctorChecksTests.cs
@@ -67,6 +67,52 @@ public class DoctorChecksTests : IDisposable
     }
 
     [Fact]
+    public async Task InstallationCheck_LegacyDotnetToolPresent_ReportsError()
+    {
+        var original = TendrilInstallHelper.UserProfileOverride;
+        TendrilInstallHelper.UserProfileOverride = _tempDir.Path;
+
+        try
+        {
+            var storeDir = Path.Combine(_tempDir.Path, ".dotnet", "tools", ".store", "ivy.tendril", "1.0.0");
+            Directory.CreateDirectory(storeDir);
+
+            var check = new InstallationCheck();
+            var result = await check.RunAsync();
+
+            var legacyStatus = result.Statuses.FirstOrDefault(s => s.Label == "Legacy .NET tool");
+            Assert.NotNull(legacyStatus);
+            Assert.Equal(StatusKind.Error, legacyStatus.Kind);
+            Assert.True(result.HasErrors);
+        }
+        finally
+        {
+            TendrilInstallHelper.UserProfileOverride = original;
+        }
+    }
+
+    [Fact]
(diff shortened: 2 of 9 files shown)


artem-ivy-ai:
### Staging removed

PR #1774, opened 2026-07-25
[00069] Persist In Flight Jobs So Status Updates Survive Master Restart

### LINKED ISSUE #1759: Best-effort job status updates fail with HTTP 404 (Not Found) when server restarts
## Description
During long-running jobs, status updates reported via `tendril job status` fail with an HTTP 404 (Not Found) error if the master server process restarts during execution.

This is because `JobService.cs` maintains active jobs in an in-memory dictionary `_jobs`. If the process restarts, this dictionary starts empty. Even though the job metadata is preserved in the SQLite database, the CLI status command receives a 404 because the job is not found in memory.

## Log Output / Evidence
From the agent's stderr log:
```
tendril job status 00633 --message="Parsing task and researching upgrade check code..."
Exit code 1
Error: Response status code does not indicate success: 404 (Not Found).
```

In `JobService.cs`:
```csharp
public bool UpdateJobStatus(string id, string message, string? planId = null, string? planTitle = null)
{
    if (!_jobs.TryGetValue(id, out var job))
        return false; // Returns false, causing the controller to respond with 404
    ...
}
```

## Proposed Solution
- Update `UpdateJobStatus` to check the database (`_database.GetJobById(id)`) if the job is not found in the active `_jobs` dictionary.
- Alternatively, persist/re-load active running jobs back into the memory dictionary upon server startup.

rorychatt:
Fixes #1759

# Summary

`tendril job status <id>` returned HTTP 404 and exit 1 whenever the Tendril master process
restarted mid-job, because `JobService` kept active jobs only in its in-memory `_jobs` dictionary.
The agent's next status report hit a server that had never heard of the job, the non-zero exit
aborted the promptware, and the job was lost. In-flight jobs are now persisted, reports fall back
to the database, and the CLI no longer treats a failed report as fatal.

## Changes

- **Migration 017** adds `ProcessId`, `ReportedPlanId`, `ReportedPlanTitle` and
  `ReportedFailureReason` to the `Jobs` table. It is picked up by `DatabaseMigrator`'s reflection
  scan, so no registration list needed updating.
- **Jobs are persisted while in flight**, not just on completion: at Pending (registration),
  Blocked, Queued, and again in `LaunchJob` once the process is up, so `Running` /`StartedAt` /
  `SessionId` / `ProcessId` / `WorkingDirectory` / `CliCommand` all reach disk.
- **`GetRecentJobs` no longer drops in-flight rows.** SQLite sorts NULLs last under
  `ORDER BY CompletedAt DESC`, so rows with a NULL `CompletedAt` — exactly the running jobs —
  were the first casualties of `LIMIT 100`. Ordering now puts them first.
- **`UpdateJobStatus` and `ReportJobFailure` fall back to the database** through a new
  `ResolveJobForReport` helper, mirroring what `GetJob` already did. A found row is re-registered
  in `_jobs` (via `TryAdd`, so a concurrent report can't create a second instance), mutated,
  persisted, and pushed to the UI. Genuinely unknown ids still return `false` → 404.
- **Startup reconciliation.** `LoadHistoricalJobs` now inspects each restored Pending/Queued/
  Running row: if its `ProcessId` is still alive the job stays Running and is flagged `Detached`;
  otherwise it becomes `Failed` with `"Interrupted by Tendril master restart"` so it stops
  showing as Running forever. `Blocked` rows are left to the existing blocked-job timer.
- **Detached jobs don't get reaped.** A restored job has no `Process` handle and no output stream
  in this process, so `RunStuckJobCheck` would see stale output and kill it. The reconcile resets
  `LastOutputAt`, and each subsequent `UpdateJobStatus` for a handle-less job refreshes it — the
  agent's own status reports act as the heartbeat.
- **The CLI is best-effort.** New `MasterClient.TryPutJson` returns `(bool Ok, string? Error)`
  instead of throwing; `job status` and `job fail` warn on stderr and exit 0. `PutJson` is
  untouched for the commands that must fail loudly.

## API Changes

- `MasterClient.TryPutJson(string relativePath, object payload, CancellationToken)` — new;
  non-throwing sibling of `PutJson`, which is unchanged.
- `JobItem.Detached` — new transient `[JsonIgnore]` bool. Not a database column, not serialised.
- `Jobs` table gains four nullable columns (migration 017). Existing databases upgrade in place;
  older Tendril builds cannot open a v17 database, which is the normal migration contract.
- `tendril job status` / `tendril job fail` now exit 0 when the master is unreachable or returns
  an error, printing a warning to stderr. Scripts that relied on a non-zero exit to detect an
  unreachable master will no longer see one.

## Files Modified

| File | Change |
| --- | --- |
| `src/Ivy.Tendril/Database/Migrations/Migration_017_JobsInFlightFields.cs` | New — four columns, `user_version = 17` |
| `src/Ivy.Tendril/Services/Plans/PlanDatabaseService.cs` | Write/read the new columns; in-flight-first `GetRecentJobs` ordering |
| `src/Ivy.Tendril/Services/Jobs/JobService.cs` | `ResolveJobForReport`, in-flight `PersistJob` calls, `ReconcileRestoredJob`, `IsAgentProcessAlive`, detached heartbeat |
| `src/Ivy.Tendril/Models/JobModels.cs` | `JobItem.Detached` |
| `src/Ivy.Tendril/Helpers/MasterClient.cs` | `TryPutJson` |
| `src/Ivy.Tendril/Commands/JobStatusCommand.cs` | Best-effort report, warn + exit 0 |
| `src/Ivy.Tendril/Commands/JobFailCommand.cs` | Best-effort report, warn + exit 0 |
| `src/Ivy.Tendril.Test/JobServiceStartupTests.cs` | 8 new tests (DB fallback, heartbeat, reconcile alive/dead/blocked) |
| `src/Ivy.Tendril.Test/PlanDatabaseServiceTests.cs` | 3 new tests (round-trip, NULLs, LIMIT) |
| `src/Ivy.Tendril.Test/DatabaseMigratorTests.cs` | Migration 017 column + insert/read test |

10 files, +562 / −9, in four commits:

- `cf38b87a` feat: persist in-flight job fields and keep them out of LIMIT's way
- `bfd5cc7d` fix: resolve agent job reports from the database after a master restart
- `7535a287` fix: make tendril job status/fail best-effort instead of fatal
- `8f150167` test: cover in-flight job persistence and restart reconciliation

## Manual Testing

Automated coverage is the primary evidence: 1802 passed / 1 skipped / 0 failed in
`Ivy.Tendril.Test`, 951 passed in `Ivy.Tendril.Agents.Test`, `dotnet format --verify-no-changes`
clean. The reconcile tests exercise a real live PID (`Environment.ProcessId`) and a real dead one
(`int.MaxValue`), and the LIMIT test seeds 1 running + 5 completed rows against real SQLite.

To confirm end to end on a live install:

1. Start the Tendril master and launch any plan job; note its id.
2. Kill the master process (leaving the agent running) and start it again.
3. From the agent's worktree run `tendril job status <id> --message="after restart"` — it should
   print `Status updated for job <id>` and exit 0, and the Jobs view should show the job still
   Running with the new message.
4. Stop the master, kill the agent process too, then start the master: the job should now appear
   as Failed with "Interrupted by Tendril master restart".
5. With the master stopped, run `tendril job status <id> --message="x"` — expect a
   `Warning: could not report status …` line on stderr and exit code 0.

## Notes

`Ivy.Tendril.slnx` references `../../../Ivy-Framework/src/Ivy/Ivy.csproj`, which does not exist on
this machine and cannot resolve from inside a worktree, so `dotnet build` on the solution fails
with MSB3202 before compiling anything. Builds were run per-project instead (the DotnetBuild
verification's documented fallback). Raised as a recommendation.

---
## Commits
- `8f150167` test: cover in-flight job persistence and restart reconciliation (plan 00069)
- `7535a287` fix: make tendril job status/fail best-effort instead of fatal (plan 00069)
- `bfd5cc7d` fix: resolve agent job reports from the database after a master restart (plan 00069)
- `cf38b87a` feat: persist in-flight job fields and keep them out of LIMITs way (plan 00069)

---

### DIFF
--- a/src/Ivy.Tendril.Test/DatabaseMigratorTests.cs
+++ b/src/Ivy.Tendril.Test/DatabaseMigratorTests.cs
@@ -460,6 +460,59 @@ public class DatabaseMigratorTests : IDisposable
         Assert.DoesNotContain("Risk", columns);
     }
 
+    [Fact]
+    public void Migration_017_JobsInFlightFields_AddsColumns()
+    {
+        new Migration_001_InitialSchema().Apply(_connection);
+        new Migration_002_Fts5Search().Apply(_connection);
+        new Migration_003_JobsTable().Apply(_connection);
+        new Migration_004_SourceUrl().Apply(_connection);
+        new Migration_005_CostsLogTimestampIndex().Apply(_connection);
+        new Migration_006_CostsCompositeIndex().Apply(_connection);
+        new Migration_007_FtsSourceUrl().Apply(_connection);
+        new Migration_008_PrStatusTable().Apply(_connection);
+        new Migration_009_JobsArgs().Apply(_connection);
+        new Migration_010_RecommendationImpactRisk().Apply(_connection);
+        new Migration_011_JobsTypedArgs().Apply(_connection);
+        new Migration_012_JobsPlanFileIndex().Apply(_connection);
+        new Migration_013_JobsWorkingDirAndCliCommand().Apply(_connection);
+        new Migration_014_JobsCleared().Apply(_connection);
+        new Migration_015_RenamePlanStates().Apply(_connection);
+        new Migration_016_DropRecommendationRisk().Apply(_connection);
+
+        Assert.Equal(16, GetUserVersion());
+
+        new Migration_017_JobsInFlightFields().Apply(_connection);
+
+        Assert.Equal(17, GetUserVersion());
+
+        var columns = new List<string>();
+        using (var pragmaCmd = _connection.CreateCommand())
+        {
+            pragmaCmd.CommandText = "PRAGMA table_info(Jobs);";
+            using var reader = pragmaCmd.ExecuteReader();
+            while (reader.Read())
+                columns.Add(reader.GetString(reader.GetOrdinal("name")));
+        }
+
+        Assert.Contains("ProcessId", columns);
+        Assert.Contains("ReportedPlanId", columns);
+        Assert.Contains("ReportedPlanTitle", columns);
+        Assert.Contains("ReportedFailureReason", columns);
+
+        using var insertCmd = _connection.CreateCommand();
+        insertCmd.CommandText = """
+                                INSERT INTO Jobs (Id, Type, PlanFile, Project, Status, Provider, ProcessId, ReportedPlanId, ReportedPlanTitle, ReportedFailureReason)
+                                VALUES ('job-1', 'ExecutePlan', 'Plan', 'Proj', 'Running', 'claude', 4242, '00069', 'Persist Jobs', 'boom')
+                                """;
+        insertCmd.ExecuteNonQuery();
+
+        using var selectCmd = _connection.CreateCommand();
+        selectCmd.CommandText =
+            "SELECT ProcessId || '|' || ReportedPlanId || '|' || ReportedPlanTitle || '|' || ReportedFailureReason FROM Jobs WHERE Id = 'job-1';";
+        Assert.Equal("4242|00069|Persist Jobs|boom", selectCmd.ExecuteScalar()?.ToString());
+    }
+
     private class FakeMigration : IMigration
     {
         private readonly List<int>? _tracker;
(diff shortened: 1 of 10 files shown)


artem-ivy-ai:
### Staging removed

PR #1859, opened 2026-08-02
[00097] Compute Duplicate Candidates at Plan Creation Instead of Job Start

rorychatt:
# Summary

## Changes

`DuplicateCandidates` was documented in the CreatePlan firmware but never computed by anything, and
the firmware told every agent that its absence proved no duplicates existed, so duplicate detection
had never run in any job. This adds a real candidate finder, prints its results at `tendril plan
create` (stdout) and `tendril plan write-revision` (stderr, exit code unchanged), and corrects the
firmware sentence in both `CreatePlan/Program.md` copies. `write-revision` is the surface that
matters: it is the only point in the CreatePlan lifecycle that runs after research, so it catches a
sibling plan that landed while the agent was working.

## API Changes

New:

- `Ivy.Tendril.Services.Plans.DuplicateCandidate` (`public sealed record`) with `FolderName`,
  `Title`, `State`.
- `Ivy.Tendril.Services.Plans.DuplicateCandidateFinder.Find(string plansDirectory, string title,
  string project, string? excludeFolderName = null)` returning
  `IReadOnlyList<DuplicateCandidate>`. Never throws.
- `DuplicateCandidateFinder.FormatBlock(IReadOnlyList<DuplicateCandidate>)` returning the
  `DuplicateCandidates:` block, or `""` for an empty list.
- `--no-duplicate-check` on `tendril plan create` and `tendril plan write-revision`
  (`PlanCreateSettings.NoDuplicateCheck`, `PlanWriteRevisionSettings.NoDuplicateCheck`).

Changed output, both additive:

- `tendril plan create` may append a fourth block after `Verifications:`. The existing
  `PlanId:` / `Directory:` / `Verifications:` blocks are byte-identical, asserted by a test.
- `tendril plan write-revision` may write a warning to **stderr**. Its stdout is still only the
  revision path, and its exit code is still 0.

No existing signature changed and nothing was removed.

## Files Modified

New:

- `src/Ivy.Tendril/Services/Plans/DuplicateCandidateFinder.cs`
- `src/Ivy.Tendril.Test/Services/DuplicateCandidateFinderTests.cs` (12 tests)
- `src/Ivy.Tendril.Test/Commands/PlanCreateDuplicateCandidatesTests.cs` (9 tests)

Changed:

- `src/Ivy.Tendril/Commands/PlanCreateCommand.cs`, `PlanWriteRevisionCommand.cs`
- `src/Ivy.Tendril/Promptwares/CreatePlan/Program.md` and the
  `src/Ivy.Tendril.TeamIvyConfig/` copy

## Manual Testing

The `Program.md` change only takes effect for agents once the promptware is redeployed:
`PromptwareDeployer.NeedsUpdate` compares a version file against the assembly version, so the
deployed copy at `<TendrilHome>/Promptwares/CreatePlan/Program.md` still contains the old sentence
until a version bump ships. To confirm the source is correct now:
`grep -c "no potential duplicates were found" src/Ivy.Tendril/Promptwares/CreatePlan/Program.md`
returns 0.

To see the CLI behavior:

1. Make a scratch plans directory with one plan whose title overlaps what you are about to create,
   e.g. `<tmp>/Plans/00063-Something/plan.yaml` containing `state: Completed`, `project:
   Ivy-Tendril`, and `title: Add the Missing Cargo Fmt Pre-Commit Guard From Plan 00042`.
2. `tendril plan create "Deliver the Blocked Cargo Fmt Pre-Commit Hook and Correct Plan 00042
   Motivation" "Ivy-Tendril" --plans-dir="<tmp>/Plans"`.
   Observe a `DuplicateCandidates:` block **after** the `Verifications:` list, one
   `folderName|title|state` line per match. Re-run with `--no-duplicate-check` and observe it is
   gone while the first three blocks are unchanged.
3. `TENDRIL_PLANS="<tmp>/Plans" tendril plan write-revision <the new id> --file some.md`.
   Observe on **stderr**: `warning: 1 possible duplicate plan(s) found. Review before this plan is
   executed:` followed by the same block. Confirm the exit code is 0, stdout is just the revision
   path, and `Revisions/001.md` exists: the warning must never block the write. The new plan must
   not list itself.

- **Test scope widened.** The plan's filter named only the two new test classes. It omits
  `PlanCliCommandTests`, the existing suite that drives both commands I changed, so passing it would
  not have shown the additive output left the existing three blocks intact. I ran the plan's filter
  as written (61 passed) and then the whole unit project (1975 passed, 1 skipped, 0 failed).
- **One extra test, beyond the plan.** `ScanPlansSubstringSearch_TheFourRealSiblings_FindsNoneOfThem`
  asserts the *old* mechanism (`plan list --search`) finds zero of the four siblings on the same
  fixture data the new finder matches all four of. Without it, the regression test passes for two
  indistinguishable reasons.
- **The plan predicted 3 siblings; the finder returns 4.** The fourth is a real match, so the test
  asserts 4. Worth a reviewer's attention since the plan's prose says three.
- **Two pre-existing em dashes replaced on an edited line.** `AGENTS.md:17` forbids U+2014. Editing
  the middle of one long bullet made the whole line an added line, carrying two em dashes that were
  already there; both became commas. The surrounding files still contain 45 and 37 on untouched
  lines, deliberately left alone to keep the diff reviewable.

## Reflection

Three new ExecutePlan memories, each verified against this worktree rather than written from
recollection: `slnx-escapes-the-worktree` (`Ivy.Tendril.slnx` references
`../../../Ivy-Framework/src/Ivy/Ivy.csproj`, which resolves outside a plan worktree and makes
`dotnet format` fail with an undiagnosable "Restore operation failed"; reproduced MSB3202 verbatim
and confirmed the per-csproj form exits 0), `added-lines-carry-old-em-dashes`, and
`measure-heuristics-on-the-real-corpus`. No existing memory was contradicted, so none was deleted;
seven were used and re-confirmed.

Re-running the corpus measurement during reflection caught a stale number: the figures in the "rank
duplicate candidates" recommendation moved because plan **00146** ("Cover the Pre-Commit Hook's
Frontend Branch in the Real Checkout") was created at 22:54 UTC today, mid-execution. It is a sixth
member of the same pre-commit cluster this plan exists because of, which is direct evidence for the
premise: a check that runs at job start cannot see a sibling that does not exist yet. The
recommendation now cites `Artifacts/measure-rate.py` instead of hard-coded constants.

---

## Commits

- 37cd9fed feat(plans): add DuplicateCandidateFinder for significant-word plan title overlap
- b041d08a feat(cli): surface DuplicateCandidates at plan create and plan write-revision
- d47feac8 fix(promptware): stop telling CreatePlan that an absent DuplicateCandidates proves no duplicates

---

### DIFF
--- a/src/Ivy.Tendril.TeamIvyConfig/Promptwares/CreatePlan/Program.md
+++ b/src/Ivy.Tendril.TeamIvyConfig/Promptwares/CreatePlan/Program.md
@@ -53,7 +53,15 @@ Do NOT read or modify `.counter` directly. Plan IDs are allocated by the `tendri
 
 ### 3. Research
 
-- **Check for duplicate plans** first — **unless `Force: true` is set in the firmware header**, in which case skip duplicate detection entirely. However, if you discover related existing plans during research (e.g., from grep results or memory), still link them via `--related-plan` on `tendril plan create`. `Force` means "create the plan regardless" — not "ignore prior work." Check the `DuplicateCandidates` firmware value. If present, it contains pre-computed matches (format: `folderName|title|state` per line). For each match, perform **state-aware duplicate detection** on those specific plans only. If `DuplicateCandidates` is absent, no potential duplicates were found — skip duplicate detection. When matches are found, decide as follows:
+- **Check for duplicate plans** first, **unless `Force: true` is set in the firmware header**, in which case skip duplicate detection entirely. However, if you discover related existing plans during research (e.g., from grep results or memory), still link them via `--related-plan` on `tendril plan create`. `Force` means "create the plan regardless", not "ignore prior work." Check the `DuplicateCandidates` firmware value. If present, it contains pre-computed matches (format: `folderName|title|state` per line). For each match, perform **state-aware duplicate detection** on those specific plans only.
+
+  An **absent** `DuplicateCandidates` header means no pre-computed match was supplied, **not** that none exists. Plans are created in concurrent batches (26 CreatePlan jobs were in flight in the batch that produced four plans for one deliverable), so a sibling covering your task may land while you research. **Always search yourself before writing the revision**, for two or three key terms from the task:
+
+  ```bash
+  tendril plan list --search "<key term>" --project="<project>"
+  ```
+
+  `tendril plan create` and `tendril plan write-revision` also print a `DuplicateCandidates:` block of their own when they find candidates (see Step 4.2). Treat every entry from any of these three sources the same way. When matches are found, decide as follows:
 
   #### Step 1: Read existing plan state
   
@@ -192,6 +200,13 @@ Verifications:
 
 Parse `PlanId` and `Directory` from the output — use these for all subsequent operations.
 
+If plans in the same project have overlapping titles, a fourth block follows the three above:
+```
+DuplicateCandidates:
+<folderName>|<title>|<state>
+```
+This is advisory and never changes the exit code. Apply the Step 3 state table to each entry before you finalize the plan (see Section 4.2 for what to do once the plan folder exists).
+
 Include optional flags as needed (always use the `--option=value` form):
 - `--source-url="<url>"` — if a source URL was extracted in Step 1
 - `--related-plan="<folder-name>"` — for each plan referenced via `[number]` syntax in the task description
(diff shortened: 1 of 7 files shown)


artem-ivy-ai:
### Staging removed

PR #1860, opened 2026-08-02
[00103] Route a Failed PreExecution to Failed Instead of Laundering It Into Completed

rorychatt:
# Summary: Route a Failed PreExecution to Failed Instead of Laundering It Into Completed

Plan 00103, project Ivy-Tendril, base branch `development`, job 01301.

## What was wrong

An agent that fails ExecutePlan step 1.7 writes `Verification/PreExecution.md` with `result: Fail` and stops.
Nothing in the server ever read that file: `grep -rIn PreExecution src/ --include=*.cs` returned exactly one
hit, the `continue` at `PlanArtifactSyncer.cs:66` that deliberately skips it. So the sole state decision,
`JobCompletionHandler.EnsurePlanStateTransitioned`, keyed only on `plan.yaml` verification rows.

Plan 00041 hit the worst path. The firmware told it to "set every verification to `Skipped`", which made
`hasIncomplete` false and routed a plan that implemented nothing to `Review`. From there the Review app,
seeing zero commits, offers a bare "Complete Plan" button, so one click made it read as a delivered plan.
That is how the `widget_harness.rs:409` fmt diff later got misattributed to 00043's scope. Plan 00043 escaped
only by accident: it left its rows `Pending`, so it landed in `Failed`.

## What changed

Four commits, 12 files, +545/-10.

### `79fc7f26` fix(plans): route a failed PreExecution to Failed instead of Review

- `PlanYamlHelper.ReadPreExecutionResult` reads the report straight from disk, reusing the existing
  `ParseVerificationResultFromReport` so both the current `---\nresult: Fail\n---` frontmatter and the legacy
  `- **Result:** Fail` body parse. Returns null for absent or unparseable, which falls through to today's
  behaviour.
- `EnsurePlanStateTransitioned` now routes to `Failed` when the report reads `Fail`, regardless of the
  verification rows. `Failed` rather than `Skipped`: it keeps the plan in the human Review queue, retains the
  worktree for inspection, and does not hide the diagnosis the agent just wrote.
- `PlanReaderService.TransitionState` refuses `Completed` when the report reads `Fail` **and** the plan has no
  commits and no PRs, throwing the new `PlanTransitionBlockedException`. The commits-and-PRs conjunct is what
  keeps config-only plans and any plan that did real work out of the block.
- The three fire-and-forget UI call sites (two in `Review/ContentView.cs`, one in `Drafts/ActionBarView.cs`)
  catch it, toast the reason, and return before `refreshPlans()`. Without that the throw would vanish into the
  handler and read as a silent no-op.
- `PlanArtifactSyncer.cs:66` was left exactly as it was: `PreExecution` still must not be synced into a
  `plan.yaml` row.

### `fa5cf106` docs(execute-plan): make the "fail the plan" instruction executable

The firmware said "fail the plan" while step 9 forbade the only verb that would do it, so the instruction had
no mechanism. Both in-repo copies of `ExecutePlan/Program.md` now spell out the contract: write the report with
`result: Fail`, call `tendril job fail`, `exit 1`. The instruction to blanket-`Skipped` the verifications is
deleted and replaced with a bolded paragraph explaining that it is precisely what routed 00041 to `Review`.
Step 9's prohibition is untouched, with one sentence added saying a failed pre-execution is signalled by
`job fail` plus a non-zero exit, never by writing `state`.

### `7f94e177` feat(doctor): surface Completed plans that failed pre-execution

`DoctorCommand.FindLaunderedCompletedPlans` lists every `Completed` plan whose `PreExecution.md` reads `Fail`,
printing the report path and a `tendril plan set <id> state Skipped` suggestion. Report only, never
auto-mutate, and the exit code is unchanged: these are historical records and retiring them is the user's call.

### `59bd0386` test(plans): cover pre-execution failure routing and the Completed block

Eleven tests. Ten in the new `PreExecutionFailureStateTests.cs` (the 00041 regression, the 00043 shape, and
negative controls for absent/Pass/Skipped reports; plus the four `TransitionState` cases including the
config-only false-positive control and the Skipped escape hatch), one pinning that `PlanValidationService`
still accepts `Completed` with empty commits and prs, one for the doctor sweep. Tests 1 through 5 drive the
real handler against a real `PlanReaderService`, not a reimplementation of the decision.

## Why the task's proposed alternative was not implemented

The task offered a fallback: have `tendril plan validate` flag `Completed` with empty `commits` and empty `prs`
as invalid. Measured on the real plans directory, 00048, 00050 and 00053 all have exactly that shape and are
healthy config-only plans (they edit verification prompts via `tendril verification set`, touch no repo files,
and all have `PreExecution: Pass`). The rule would call three good plans invalid, and since `Validate` runs
from `WritePlan` on every plan write it would make them unwritable, not merely flagged. The discriminator that
separates the groups cleanly is the execution record: `PreExecution: Fail` on both phantoms, `Pass` on all
three innocents. Test 10 exists so nobody re-adds the rejected rule.

## Verification

| Verification | Result | Notes |
| --- | --- | --- |
| PreExecution | Pass | Zero validation blocks in the revision, so step 1.7 skipped validation. Premise claims independently confirmed |
| DotnetFormat | Pass | `--verify-no-changes` on both csprojs, exit 0, nothing rewritten |
| DotnetBuild | Pass | `Ivy.Tendril` and `Ivy.Tendril.Test`, both `0 Warning(s) 0 Error(s)` under `-warnaserror` |
| DotnetTest | Pass | Plan scope 115/115; full `Ivy.Tendril.Test` 1966 passed, 1 self-skip, 0 failed |
| CheckResult | Pass | Every deliverable checked, including the doctor CLI exercised once with output and exit code |
| WidgetsFrontendCheck | Skipped | Pre-set in plan.yaml, no frontend changes |
| WidgetsFrontendTest | Skipped | Same |

The plan required proving the guard is load-bearing. With the `preExecution == VerificationStatus.Fail` term
removed, `EnsurePlanStateTransitioned_PreExecutionFail_AllVerificationsSkipped_SetsFailed` fails with
`Expected: "Failed" / Actual: "Review"`, the literal 00041 bug, while the Pending test still passes. The term
was restored and all 10 tests pass.

**The two `ExecutePlan/Program.md` copies are not the same file.** The plan asserts they have the "same
content" at listed line numbers, and that the `src/Ivy.Tendril` copy is byte-identical to the deployed
`<TendrilHome>` one. `diff` disproves it: the `Ivy.Tendril.TeamIvyConfig` copy has a "Time Budget Awareness"
section the src copy lacks, lacks the src copy's status-report lines, and still describes the older manual
worktree flow. All the plan's cited line numbers are therefore wrong for at least one copy.

I made all three edits to both copies by matching on content rather than line number, then confirmed each with
the greps the plan prescribed. Same deliverable, different navigation. Anyone editing these files later should
not trust cross-copy line numbers.

**Test count and grouping.** The plan listed 11 tests; I wrote 12. The extra one,
`TransitionState_ToSkipped_PreExecutionFail_IsAllowed`, exercises the retire path the block message tells the
user to take. A block whose stated escape hatch is untested is a block that can strand a plan. The plan also
placed `Doctor_ListsCompletedPlansWithFailedPreExecution` in the new file; it went into
`DoctorCommandPlansTests.cs` instead, next to the `FindPruneCandidates` tests whose fixture helpers it reuses,
matching the plan's own instruction to "Follow DoctorCommandPlansTests.cs". The tests were committed as a
fourth commit rather than folded into the three, so the guard change and its proof are separable in history.

**`EnsurePlanStateTransitioned` visibility.** Changed from `private` to `internal` so the tests drive the real
decision point. `Ivy.Tendril.csproj` already has `InternalsVisibleTo Include="Ivy.Tendril.Test"`, and the class
itself is `internal`, so nothing new is exposed outside the assembly.

## Known issues I did not fix

Two failures inherited from `development`, both outside this plan's scope and both reproduced on a clean
checkout of the base branch:

- `Ivy.Tendril.Agents.Test` fails to build under `-warnaserror`: `AntigravityEventParserTests.cs(38,25):
  error CS8602`. Builds clean without the flag.
- Two of its tests fail on `development`: `AntigravityCliTests.BuildProcessSpec_BasicInvocation_...` and
  `OpenCodeModelCatalogTests.GetStaticModels_ReturnsDefault` (asserts `Single()` against a catalog that now has
  8 models).

This plan changed no file in that project (`git diff --stat development...HEAD -- src/Ivy.Tendril.Agents.Test/`
is empty). Both are filed as recommendations rather than fixed here, to keep this diff to the state-routing fix.

## Plan 00090 interaction

Plan 00090 guards the same `TransitionState` chokepoint for a different condition (a verification row in
`Fail`). It has not landed: neither its `PlanTransitionBlockedException` nor its `GetFailedVerifications` was
present on `development`, so this plan introduced the exception. Whoever lands second must rebase and
**compose** the two conditions rather than replacing one; the exception's docstring says so. The two are
genuinely complementary: 00041 has no `Fail` row at all, so 00090's check would pass it straight through.

---

## Commits

- 79fc7f26 fix(plans): route a failed PreExecution to Failed instead of Review
- fa5cf106 docs(execute-plan): make the "fail the plan" instruction executable
- 7f94e177 feat(doctor): surface Completed plans that failed pre-execution
- 59bd0386 test(plans): cover pre-execution failure routing and the Completed block

---

### DIFF
--- a/src/Ivy.Tendril.TeamIvyConfig/Promptwares/ExecutePlan/Program.md
+++ b/src/Ivy.Tendril.TeamIvyConfig/Promptwares/ExecutePlan/Program.md
@@ -118,7 +118,16 @@ After reading the plan revision, scan it for code validation markers to detect s
 3. **Decision logic:**
    - **If no validation blocks found** → Skip validation, proceed to worktree creation (backward compatible)
    - **If all validation blocks pass** → Proceed to worktree creation
-   - **If any validation fails** → Fail the plan immediately with a detailed report
+   - **If any validation fails** → Stop here and fail the job. "Fail the plan" is a concrete
+     three-step contract, not a state change:
+     1. Write `<TendrilPlanFolder>/Verification/PreExecution.md` with `result: Fail` and the detailed
+        per-block report below.
+     2. Call `tendril job fail TendrilJobId --message="PreExecution failed: <reason>"`.
+     3. `exit 1`.
+
+     The server reads that report and routes the plan to `Failed` on its own, so do **not** set the
+     state yourself (see step 9). Do not proceed to worktree creation, do not run verifications, and
+     do not commit anything: there is nothing to verify on an unmodified base branch.
 
 4. **Write validation report** — Create `<TendrilPlanFolder>/Verification/PreExecution.md`:
 
@@ -151,7 +160,17 @@ date: <CurrentTime>
    - A `<details><summary>Still relevant?</summary>` block whose body starts with `No.`
    - Phrases like *"Already applied"*, *"This plan is redundant"*, *"This plan is superseded"*, or *"previously attempted … was merged to main via PR #NNNN"* in the `## Problem` or `## Solution` sections.
 
-   If any marker is found, verify the claim: run `gh pr view <cited PR> --json state,mergeCommit` (must be `MERGED`), confirm the cited commit is in `git log origin/<default-branch>`, and byte-compare the plan's proposed code against the current file contents. If all three checks pass, write `Verification/PreExecution.md` with `Result: Fail`, write `Artifacts/summary.md` documenting the no-op, set every verification to `Skipped` via `tendril plan set-verification <plan-id> <name> Skipped`, and fail the plan **without creating a worktree** — running verifications on unchanged code wastes the time budget and produces a 0-commit PR that CreatePr cannot process.
+   If any marker is found, verify the claim: run `gh pr view <cited PR> --json state,mergeCommit` (must be `MERGED`), confirm the cited commit is in `git log origin/<default-branch>`, and byte-compare the plan's proposed code against the current file contents. If all three checks pass, fail **without creating a worktree** (running verifications on unchanged code wastes the time budget and produces a 0-commit PR that CreatePr cannot process):
+
+   1. Write `Verification/PreExecution.md` with `result: Fail` and the evidence for each of the three checks.
+   2. Write `Artifacts/summary.md` documenting the no-op.
+   3. Call `tendril job fail TendrilJobId --message="PreExecution failed: <reason>"` and `exit 1`.
+
+   **Do not set the verifications to `Skipped`.** An earlier revision of this document told you to, and
+   that is exactly how a never-executed plan reached `Review`: the server's state decision keys on
+   verification rows, so blanket `Skipped` makes every row look complete and routes the plan to
+   `Review`, where one click marks it `Completed` with zero commits. Leave the rows `Pending` and let
+   the `PreExecution: Fail` report and the non-zero exit do the work.
 
 ### 2. Create Worktrees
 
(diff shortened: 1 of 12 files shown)


artem-ivy-ai:
### Staging removed

PR #1861, opened 2026-08-02
[00090] Block Review to Completed When a Verification Failed and Surface Partial Delivery to Duplicate Detection

rorychatt:
# Plan 00090 Execution Summary

Block Review to Completed when a verification failed, and surface partial delivery to duplicate
detection.

## Changes

Plan 00042 sits at `state: Completed` with a merged PR while its own `CheckResult` verification
records `Fail` and says the primary deliverable is missing. Four later plans re-derived the same
missing work because 00042 read as done. This plan fixes both halves of that defect in four commits.

**Commit 1 (`60e3c8d2`) blocks the transition at a single chokepoint.** `PlanReaderService.TransitionState`
now throws `PlanTransitionBlockedException` when asked for `Completed` while any verification is
`Fail`. Because the CLI, both Review app buttons and the Drafts menu item all funnel through it, one
check covers them all. The failed names are read from plan.yaml on disk rather than the database, so a
stale cache cannot green-light a blocked transition. `Pending` is deliberately not included:
`JobCompletionHandler.EnsurePlanStateTransitioned` already routes Pending to `Failed`, and only `Fail`
is a positive statement that a gate ran and rejected the work.

The three UI call sites run in fire-and-forget event handlers, where an unhandled throw is a silent
no-op, so they go through `PlanCompletionAction.TryComplete`. The Review app's optimistic path skips
its `refreshPlans()` and its background worktree cleanup on a block rather than pretending the plan is
done. Review offers `PartialDeliveryDialog` as an override; the Drafts menu, which has no override
affordance, toasts the failed gate names.

**Commit 2 (`27ee2f33`) adds the escape hatch and makes CreatePr verification-aware.** A hard block
with no way out is worse than the bug, so `tendril plan set <id> state Completed` gained
`--allow-failed-verifications`, and the exception message names it. CreatePr step 6 now reads the
verifications before setting state: `add-pr` stays unconditional because recording a PR is always
correct, but a `Fail` leaves the plan in `Failed` and reports the gate. The step's MANDATORY-CLOSEOUT
framing is preserved; a failed gate is now a legitimate non-`Completed` outcome, not a skipped step.

**Commit 3 (`fcb31e1c`) makes partial delivery visible.** Blocking only protects plans completed from
now on; 00042 is already `Completed` on disk and the duplicate-detection table keys on `state` alone.
`PlanYaml.PartialDelivery` is set whenever an override is used, surfaced in the Review UI next to the
title (which is what readers trust), and readable via `tendril plan get <id> partialDelivery`. The
CreatePlan state table gained a row that overrides the "otherwise trash" verdict for these plans, plus
two mechanical greps so it is not a judgement call.

Beyond the plan's four listed sites, the REST `PUT /api/plans/{planId}` endpoint and the MCP
`tendril_plan_set` tool were found to write `plan.State` directly, bypassing the guard exactly as
`plan set` did. The rule was centralized in `PlanCompletionGuard.ApplyState`, which all three direct
writers now share, each with its own explicit opt-out.

**Commit 4 (`a4bb734b`) backfills.** `tendril plan doctor` now lists every plan at `Completed` with a
`Fail` verification and no `partialDelivery` flag, and prints the command to flag them. Report only,
never mutate: these are historical records and the call is the user's. Already-flagged plans drop off
the list, so it shrinks as the user works through it.

### Notable decisions

- **No schema migration.** `PartialDelivery` is additive and absent-means-false, so
  `CurrentSchemaVersion` stays at 3. A bump would force a one-time re-sweep of every non-terminal
  plan, which is not worth it for an additive bool. The SQLite mirror derives the flag from the
  existing `YamlRaw` column rather than gaining a dedicated column, for the same reason.
- **"Any verification Fail", not "CheckResult is Fail".** The task names `CheckResult` because that is
  where 00042's failure landed, but `CheckResult` is a per-project convention and the same hole exists
  for any required gate.
- **00042 itself was left untouched.** The plan asks for a report, not an auto-fix. It is the first
  thing `tendril plan doctor` will now surface.

## API Changes

**`IPlanReaderService`** gained two members:

```csharp
IReadOnlyList<string> GetFailedVerifications(string folderName);
void CompleteWithPartialDelivery(string folderName);
```

`TransitionState(folderName, PlanStatus.Completed)` now throws `PlanTransitionBlockedException` when a
verification is `Fail`. Existing callers that pass any other state are unaffected. Five test stubs
were updated for the two new members.

**New public types** in `Ivy.Tendril.Services.Plans`: `PlanCompletionGuard` (with
`FailedVerificationNames` and `ApplyState`) and `PlanTransitionBlockedException` (carrying
`FolderName` and `FailedVerifications`). New public helper `Ivy.Tendril.Helpers.PlanCompletionAction`.

**CLI:**

```bash
tendril plan set <id> state Completed --allow-failed-verifications   # new flag
tendril plan get <id> partialDelivery                                # new readable field
tendril plan doctor                                                  # new backfill section
```

`plan set <id> state Completed` is now a breaking change for scripts that complete a plan over a
failed gate: it exits non-zero unless the new flag is passed.

**REST:** `PUT /api/plans/{planId}` returns `400` for `state: Completed` with a failed verification.
`SetFieldRequest` gained an optional `allowFailedVerifications` (defaults to `false`, so existing
request bodies still bind). `?field=partialDelivery` is readable.

**MCP:** `tendril_plan_set` gained an optional `allowFailedVerifications` parameter and returns
`Error: ...` for a blocked completion. `tendril_get_plan` accepts `field: partialDelivery`.

**plan.yaml:** new optional `partialDelivery: true` key, omitted when false. No schema version change,
and existing files are not rewritten.

## Files Modified

32 files, 896 insertions, 39 deletions.

**New (6):**

| File | Purpose |
|---|---|
| `src/Ivy.Tendril/Services/Plans/PlanCompletionGuard.cs` | the shared rule, used by all three direct plan.yaml writers |
| `src/Ivy.Tendril/Services/Plans/PlanTransitionBlockedException.cs` | carries the failed gate names; message states the escape hatch |
| `src/Ivy.Tendril/Helpers/PlanCompletionAction.cs` | shared complete-plan action for the fire-and-forget UI handlers |
| `src/Ivy.Tendril/Apps/Review/Dialogs/PartialDeliveryDialog.cs` | the override affordance |
| `src/Ivy.Tendril.Test/PlanCompletionVerificationGuardTests.cs` | 10 tests built on 00042's real verification list |
| (no new docs files; existing pages extended) | |

**Enforcement and state (5):** `Services/Plans/PlanReaderService.cs`,
`Services/Plans/IPlanReaderService.cs`, `Models/PlanModels.cs`,
`Services/Plans/PlanDatabaseService.cs`, `Commands/PlanSetCommand.cs`

**Other write and read paths (4):** `Controllers/PlanController.cs`, `Mcp/Tools/PlanTools.cs`,
`Commands/PlanGetCommand.cs`, `Commands/DoctorCommand.cs`

**UI (3):** `Apps/Review/ContentView.cs`, `Apps/Review/SidebarView.cs`, `Apps/Drafts/ActionBarView.cs`

**Promptware and prompts (5):** `Promptwares/CreatePr/Program.md` and
`Promptwares/CreatePlan/Program.md` in both `Ivy.Tendril` and `Ivy.Tendril.TeamIvyConfig`, plus
`Prompts/Plans.md`

**Docs (4):** `09_Advanced/01_CLI/00_Overview.md`, `01_CLI/01_Plan.md`, `02_REST.md`, `03_MCP.md`

**Tests (7):** `PlanCompletionVerificationGuardTests.cs` (new), `DoctorCommandPlansTests.cs`
(+4 tests), and five `IPlanReaderService` stub updates (`JobServiceCompletionGuardTests.cs`,
`JobServiceDependencyAutoRetryTests.cs`, `JobServiceRetryBlockedTests.cs`,
`JobServiceSplitPlanCompletionTests.cs`, `JobsAppPromptDisplayTests.cs`)

## Manual Testing

Automated coverage is 83/83 in the plan's declared scope across three consecutive runs, and the guard
test was proven to fail when the `TransitionState` check is reverted. Beyond that, the CLI behaviors
the plan documents were exercised against a throwaway `TENDRIL_HOME` seeded with a 00042-shaped plan
(`CheckResult: Fail`, `state: Review`), using the worktree's own build. Full transcripts with exit
codes are in `Verification/CheckResult.md`. Fixture directories were deleted afterwards.

| Check | Result |
|---|---|
| `plan set 00042 state Completed` | exit 1, message names `CheckResult` and the escape hatch |
| plan.yaml after the block | still `state: Review`, no `partialDelivery` line: the block happens before any write |
| `plan set 00042 state Completed --allow-failed-verifications` | exit 0, warning printed, `state: Completed` and `partialDelivery: true` on disk |
| `plan get 00042 partialDelivery` | `true` |
| `plan get 00042 verifications` | `Name=Status` format confirmed, so the `grep -i "=Fail"` CreatePlan now instructs works |
| `plan doctor --all` over 3 plans | surfaces only the unflagged partial; skips the already-flagged and the fully-verified |
| plan.yaml after `plan doctor` | unchanged for both non-reported plans: report-only confirmed |

Both prose greps the plan mandated were run: all six remaining `state Completed` occurrences across
the two CreatePr copies are verification-aware, and the new CreatePlan table row is present in both
copies (main `:104`, team `:67`), ordered before the pre-existing `Completed (with merged PR)` row so
the narrower case matches first.

**Not manually exercised:** the Review app UI paths (the "Complete Plan" button, the "Set Completed"
menu item, `PartialDeliveryDialog`, and the Partial Delivery badges) and the Drafts "Mark as
Completed" toast. These need a running desktop app and were verified by reading the call sites plus
the shared `PlanCompletionAction` they delegate to, whose blocked path is covered by the guard tests.
The REST and MCP opt-outs were likewise verified by inspection rather than by driving a live server.

**Whole-project test run:** two pre-existing flaky tests surface under full parallel load
(`PromptwareReadMemoryCommandTests` is missing `[Collection("TendrilHome")]`, and `PlanFileLock.Acquire`
retries only on `IOException` while Windows raises `UnauthorizedAccessException` for a delete-pending
handle). Both were reproduced with this plan's test class excluded entirely, and both involve code
byte-identical to the base branch. Analysis is in `Verification/DotnetTest.md`; both are filed as
recommendations.

---

## Commits

- 60e3c8d25115a42a17d93d36d0a0804f02db8c00 fix(plans): block Review to Completed when a verification failed
- 27ee2f3364b6e3817a03e93ed35b62204c825b62 feat(cli): add --allow-failed-verifications and make CreatePr verification-aware
- fcb31e1cc674d6013bfdb68dfcb8304b71c86790 feat(plans): surface partial delivery to duplicate detection
- a4bb734bd823095cf00abcc44753099359bdefa5 feat(doctor): report plans completed over a failed verification (plan 00090)

---

### DIFF
--- a/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/00_Overview.md
+++ b/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/00_Overview.md
@@ -92,6 +92,16 @@ Every command supports `--help` for detailed usage. For example: `tendril plan c
 
 Validates your Tendril installation — checks `TENDRIL_HOME`, `config.yaml`, required tools (`gh`, `git`), `pwsh`, database schema, and agent model availability. Always a good first step when something isn't working.
 
+#### plan doctor
+
+```terminal
+>tendril plan doctor
+>tendril plan doctor --all
+>tendril plan doctor --fix
+```
+
+Scans every plan folder and reports health: missing or malformed `plan.yaml`, stale worktrees, and plans left `Completed` over a failed verification. See [Plan](01_Plan.md) for the full option and health-code reference.
+
 #### run
 
 ```terminal
--- a/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/01_Plan.md
+++ b/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/01_Plan.md
@@ -248,3 +248,13 @@ With `--fix`: creates scaffold YAML for missing files, fills in missing fields, 
 >tendril plan doctor --fix
 >tendril plan doctor --prune
 ```
+
+### Partial delivery backfill
+
+The report also lists plans sitting at `Completed` with a verification in the `Fail` state and no `partialDelivery` flag. These predate the completion guard, so duplicate detection reads them as fully delivered even though the deliverable may be missing.
+
+Nothing is mutated: these are historical records, and whether a given plan really was a partial delivery is the user's call. Review each one, then flag the genuinely partial ones:
+
+```terminal
+>tendril plan set <id> state Completed --allow-failed-verifications
+```
--- a/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/01_Plan.md
+++ b/src/Ivy.Tendril.Docs/Docs/09_Advanced/01_CLI/01_Plan.md
@@ -81,7 +81,7 @@ Lists plans with optional filters.
 
 Prints the full YAML, or a single field value when `[field]` is provided.
 
-**Scalar fields:** `state`, `project`, `level`, `title`, `created`, `updated`, `executionProfile`, `initialPrompt`, `sourceUrl`, `priority`
+**Scalar fields:** `state`, `project`, `level`, `title`, `created`, `updated`, `executionProfile`, `initialPrompt`, `sourceUrl`, `priority`, `partialDelivery`
 
 **List fields:** `repos`, `prs`, `commits`, `verifications`, `dependsOn`, `relatedPlans`, `recommendations` (each item on its own line)
 
@@ -89,9 +89,12 @@ Prints the full YAML, or a single field value when `[field]` is provided.
 
 ```terminal
 >tendril plan set <plan-id> <field> <value>
+>tendril plan set <plan-id> state Completed --allow-failed-verifications
 ```
 
 Updates a single field and bumps the `updated` timestamp automatically.
+
+Setting `state` to `Completed` is refused while any verification is in the `Fail` state: a plan that reads as done while a gate rejected the work hides a missing deliverable from duplicate detection. Re-run the verification, or set it to `Skipped` with an explicit reason. `--allow-failed-verifications` records the transition anyway and sets `partialDelivery: true`, which marks the plan's deliverable as possibly missing.
 
 #### plan update
 
--- a/src/Ivy.Tendril.Docs/Docs/09_Advanced/02_REST.md
+++ b/src/Ivy.Tendril.Docs/Docs/09_Advanced/02_REST.md
@@ -67,6 +67,8 @@ Content-Type: application/json
 
 Supported fields: `state`, `project`, `level`, `title`, `executionProfile`, `initialPrompt`, `sourceUrl`, `priority`.
 
+Setting `state` to `Completed` returns `400` while any of the plan's verifications is in the `Fail` state. Add `"allowFailedVerifications": true` to record it anyway; the plan is then flagged with `partialDelivery: true`, which marks its deliverable as possibly missing.
+
 ### Add Repository
 
 ```
--- a/src/Ivy.Tendril.Docs/Docs/09_Advanced/03_MCP.md
+++ b/src/Ivy.Tendril.Docs/Docs/09_Advanced/03_MCP.md
@@ -38,7 +38,7 @@ All tools are prefixed with `tendril_` and provide the same capabilities as the 
 
 | Tool | Parameters | Description |
 |------|------------|-------------|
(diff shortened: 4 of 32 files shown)


artem-ivy-ai:
### Staging removed

PR #1862, opened 2026-08-02
[00121] Serialize Config Writes So a Concurrent Verification Set Cannot Silently Drop a Sibling Prompt

rorychatt:
# Plan 00121: Serialize Config Writes So a Concurrent Verification Set Cannot Silently Drop a Sibling Prompt

**Job:** 01318. **Branch:** `tendril/00121-SerializeConfigWritesSoAConcurrentVerificationSetCannotSilen`
(worktree, base `8d1ec1c3` on `development`). **All 4 in-scope verifications Pass**, 2 Skipped
(frontend, nothing touched).

## What was wrong

`tendril verification set` and every other `config.yaml` mutator was a read-modify-write with no
concurrency control. `ConfigService` loads the whole file in its constructor, the command mutates one
field, and `SaveSettings()` serializes the **entire** settings graph back. Two processes writing
different keys in the same window therefore produced a file where one write never happened, and
**both processes printed success and exited 0**. The plan measured 7 lost writes in 7 paired
iterations (job 01280). Nothing detected it, so the only mitigation was remembering to re-read every
sibling key afterwards.

## What was built

An advisory lock, per the plan's chosen option, modelled on the `PlanFileLock` the repo already uses
for the identical `plan.yaml` hazard.

| Commit | Contents |
| --- | --- |
| `0bad94a4` | `Helpers/ConfigFileLock.cs`, `ConfigService.MutateAndSave`, `IConfigService` member, 8 test doubles |
| `2cb2271d` | The 17 out-of-process `SaveSettings()` sites converted to `MutateAndSave` |
| `3b037cf0` | `ConfigFileLockTests.cs` and the `VerificationCommandTests` sibling-survival case |
| `efabd89f` | Fix: `ReloadSettings` must run `ExpandRepoPaths` (regression found by DotnetTest) |
| `5106242b` | Fix: foreign lock holder must yield `TimeoutException`, not raw `IOException` (found by CheckResult) |

`ConfigFileLock` is `configPath + ".lock"` opened `FileShare.None` + `DeleteOnClose`, retried 50 x
100ms, then a `TimeoutException` naming the path, which the CLI surfaces as `Error: ...` and exit 1.
That is the "fails loudly" property the plan wanted in place of compare-and-swap. It also holds an
in-process `SemaphoreSlim`, because `FileShare.None` does not exclude two handles in the same process
and the TUI has 21 in-process writers plus a `FileSystemWatcher` reload.

`MutateAndSave` acquires the lock, reloads, applies the callback, writes, releases. Acquisition
happens at exactly **one** site, which matters: `ReloadSettings` reaches `WriteSettingsToDisk` through
`SyncAuthFromEnvironmentAndPersistIfNeeded`, so a second acquisition inside the write would deadlock.

The 21 `Apps/` sites deliberately stay on `SaveSettings()`; converting them would let a reload stomp
unsaved dialog state, and the in-process guard is what protects them from a concurrent CLI. Both
methods now carry doc comments saying which to use, so the split reads as intentional.

## Proof it works

Against a published build in an isolated `TENDRIL_HOME`, re-running the plan's own reproduction:

```
before (plan's measurement):  7 of 7 iterations lost a write
after:                        0 of 7
```

The cross-command case the plan calls out (`project add-repo` racing `verification set`) also keeps
both writes. With the lock held by a foreign process, a write exits 1 after the full 5s budget with
`Error: Could not acquire config lock at ...config.yaml.lock after 5000ms` and leaves the config
untouched.

182/182 tests pass in the plan's filter. The regression test was proven non-vacuous, and each half of
the fix proven load-bearing, by mutation: swapping `MutateAndSave` back for load+mutate+`SaveSettings`
fails 3/3 with `Expected: "RACE-BETA" / Actual: "SEED-BETA"` (the losing write reverting to its seed,
exactly the reported bug); removing the reload fails 2; removing the lock fails 1. Every probe was
reverted and `grep -rn "MUTATION PROBE"` returns nothing.

**1. Eight `IConfigService` test doubles, not four.** The plan names
`FreshInstallConfigService`, `TestConfigService` (GitServiceTests), `StubConfigService` and
`TestPlanConfigService`. There are also `TestConfigService` in `JobServiceConcurrencyTests`,
`FakeConfigService` in `JobServiceDeletionTests`, and `FakeConfigService` in `JobServiceStartupTests`.
All 8 were updated; an interface member is not optional, so the build would fail otherwise. Flagged in
`Verification/PreExecution.md` before implementation.

**2. `SemaphoreSlim` instead of the sketched `private static readonly object InProcessLock = new()`.**
The plan's snippet implies `Monitor`, which is thread-affine. The returned `IDisposable` can be
disposed on a different thread than acquired it, which would throw
`SynchronizationLockException`. Same intent and same reference (`AllocatePlanId`), correct primitive.

**3. Three tests beyond the plan's five.** `Acquire_LockFileIsDeletedOnRelease` turns the plan's step 4
prose into an assertion; the other two are regression tests for the defects below.

**4. Two defects found during verification, fixed here rather than deferred.**

- `ReloadSettings` skipped the `ExpandRepoPaths` that the constructor's `FinalizeConfiguration`
  applies. Latent before this plan, because nothing reloaded mid-command; but `MutateAndSave` reloads
  inside the lock by design, so the callback started seeing raw backslashed repo paths and would have
  persisted them, breaking `ProjectConfig.GetRepoRef`. Caught by
  `ProjectCommandTests.RemoveRepo_NotFound_ListsAvailable`, which was only in scope because the plan
  said to add `ProjectCommandTests` to the filter if the class existed.
- The cross-process lock path threw a raw `IOException`. The retry loop's `when (i < maxRetries - 1)`
  filter stops catching on the last attempt, so the trailing `throw new TimeoutException` was
  unreachable for a foreign holder. Only the in-process branch produced a `TimeoutException`, and the
  plan's `Acquire_TimesOutLoudly` goes through exactly that branch, so the suite could not catch it.
  Found only by exercising the CLI end to end, as CheckResult step 6 requires. `PlanFileLock` has the
  same unreachable throw (filed as a recommendation, not changed here).

## Not done, deliberately

- `Ivy.Tendril.Agents.Test` does not satisfy `-warnaserror` on `development` (pre-existing CS8602 at
  `AntigravityEventParserTests.cs:38`). Proven pre-existing rather than assumed: last touched by
  `45874d50`, an ancestor of `development`; the file is byte-identical to `development`; this plan
  changes no file in that project; and `Ivy.Tendril.Agents` has no ProjectReference to `Ivy.Tendril`,
  so it cannot see any of these changes. Left alone per the DotnetBuild prompt's "warnings introduced
  by the plan", filed as a recommendation.
- `ConfigService.SetTendrilHome` is a third load path that also skips `ExpandRepoPaths`. It is
  `internal` and test-only, not on the `MutateAndSave` path. Filed as a recommendation.
- No PR was opened and the plan state was not changed, per the ExecutePlan rules. The worktree is
  clean and left in place.

## Artifacts

- `Verification/PreExecution.md`, `DotnetFormat.md`, `DotnetBuild.md`, `DotnetTest.md`, `CheckResult.md`
- `Artifacts/recommendations.md`

---

## Commits
- 5106242b Throw TimeoutException when a foreign holder owns the config lock (plan 00121)
- efabd89f Expand repo paths on reload so MutateAndSave keeps them normalized (plan 00121)
- 3b037cf0 Add ConfigFileLock and concurrent-writer tests (plan 00121)
- 2cb2271d Convert the 17 out-of-process config mutators to MutateAndSave (plan 00121)
- 0bad94a4 Add ConfigFileLock and ConfigService.MutateAndSave (plan 00121)

---

### DIFF
--- a/src/Ivy.Tendril.Test/BackgroundServiceActivatorTests.cs
+++ b/src/Ivy.Tendril.Test/BackgroundServiceActivatorTests.cs
@@ -275,6 +275,11 @@ public class BackgroundServiceActivatorTests : IAsyncLifetime
         {
         }
 
+        public void MutateAndSave(Action<TendrilSettings> mutate)
+        {
+            mutate(Settings);
+        }
+
         public void ReloadSettings()
         {
         }
--- a/src/Ivy.Tendril.Test/ConfigFileLockTests.cs
+++ b/src/Ivy.Tendril.Test/ConfigFileLockTests.cs
@@ -21,7 +21,10 @@ public class ConfigFileLockTests : IDisposable
         Environment.SetEnvironmentVariable("TENDRIL_HOME", _tempDir.Path);
 
         var yaml = @"
-projects: []
+projects:
+  - name: PathProject
+    repos:
+      - path: D:\Repos\Test
 verifications:
   - name: Alpha
     prompt: SEED-ALPHA
@@ -149,6 +152,26 @@ verifications:
         var reloaded = new ConfigService();
         Assert.Equal("OUT-OF-BAND", PromptOf(reloaded.Settings, "Alpha"));
         Assert.Equal("LATER-BETA", PromptOf(reloaded.Settings, "Beta"));
+    }
+
+    /// <summary>
+    ///     MutateAndSave reloads inside the lock, and ReloadSettings used to skip the ExpandRepoPaths
+    ///     that the constructor's load path applies. The callback therefore saw raw backslashed repo
+    ///     paths and the write persisted them in that form, which broke ProjectConfig.GetRepoRef for
+    ///     `project remove-repo` and friends. Both reads must agree.
+    /// </summary>
+    [Fact]
+    public void MutateAndSave_KeepsRepoPathsNormalized()
+    {
+        var config = new ConfigService();
+        var asLoaded = config.Settings.Projects.Single().Repos.Single().Path;
+        Assert.DoesNotContain('\\', asLoaded);
+
+        string? observed = null;
+        config.MutateAndSave(s => observed = s.Projects.Single().Repos.Single().Path);
+
+        Assert.Equal(asLoaded, observed);
+        Assert.Equal(asLoaded, new ConfigService().Settings.Projects.Single().Repos.Single().Path);
     }
 
     /// <summary>The lock file must not be mistaken for config, and must not survive the lock.</summary>
--- a/src/Ivy.Tendril.Test/ConfigFileLockTests.cs
+++ b/src/Ivy.Tendril.Test/ConfigFileLockTests.cs
@@ -116,6 +116,29 @@ verifications:
         Assert.Contains(ConfigPath + ".lock", captured!.Message);
     }
 
+    /// <summary>
+    ///     Acquire_TimesOutLoudly exercises the in-process semaphore, which is not the branch a real
+    ///     race takes. Here the lock file is held by a foreign handle, so the FileStream keeps throwing
+    ///     IOException until the budget runs out. That must surface as the same TimeoutException naming
+    ///     the lock path, not as a raw "cannot access the file" sharing violation.
+    /// </summary>
+    [Fact]
+    public void Acquire_ForeignHolder_TimesOutWithLockPath()
+    {
+        var lockPath = ConfigPath + ".lock";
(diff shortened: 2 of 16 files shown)


artem-ivy-ai:
### Staging removed

PR #1983, opened 2026-08-08
fix(config): prevent test runs from polluting live TENDRIL_HOME config

rorychatt:
This PR fixes PathHelper.GetDefaultTendrilHome() so that an explicitly specified TENDRIL_HOME environment variable is respected directly instead of falling back to user-level tendril home when config.yaml is not yet present, preventing test runs from overwriting live config.yaml files.


rorychatt:
@claude review this

artem-ivy-ai:
### Staging removed

PR #1995, opened 2026-08-10
fix(windows): wrap CLI execution in cmd.exe /S /c in HealthCheckRunner

rorychatt:
Fixes an issue on Windows where Onboarding step 1 gets stuck on 'Setting Up Claude' because HealthCheckRunner launches .cmd batch scripts directly without cmd.exe /S /c, causing stdio stream readers to hang indefinitely.


artem-ivy-ai:
### Staging removed

PR #2047, opened 2026-08-15
Optimize Review and Drafts app performance and eliminate tab switching lag

rorychatt:
### Summary of Changes
- **Async Git Tab Computation**: Moved `GitTabDataBuilder.BuildGitTabData` into the background `planContentQuery` instead of executing synchronous git CLI processes on the UI thread during every `Build()` render.
- **Lazy Tab Content Rendering**: Tab contents (such as `ChangesTabView` with multi-file `PlanDiffView` widgets, `GitTabView`, `ArtifactsTabView`, etc.) are now rendered only when their respective tab is selected. This prevents constructing and serializing all git diffs across all tabs on every state change or plan switch.
- **Precomputed File Diffs**: Added `FileDiffs` to `AllChangesData` and pre-split the diff in the background thread when plan changes load, avoiding regex parsing on the UI thread.

### Verification
- All 2,183 unit tests in `Ivy.Tendril.Test` passed.


artem-ivy-ai:
### Staging removed

PR #2049, opened 2026-08-15
Prevent stale output timeout on processes that already exited

rorychatt:
## Summary
- Added check in `JobMonitor.RunStaleOutputWatchdog` to avoid triggering stale output timeout if `job.Process.HasExited` is true.
- Ensures jobs whose process has exited cleanly complete via normal exit handling.


artem-ivy-ai:
### Staging removed

PR #2128, opened 2026-08-18
[00093] Stop the Worktree Reaper Orphaning Unpushed Plan Commits

nielsbosma:
# Summary

## Changes

Removed `Review` from `WorktreeCleanupService.StaleReapStates`, so the background worktree reaper no
longer deletes a `Review` plan's worktree — and, critically, no longer runs `git branch -D` on the
plan's branch, which was the only ref holding the commits execution produced. `Review` now behaves
like the active/transient states: `ResolveGrace` returns `null` for it and `CleanupPlanWorktrees`
returns early, at any idle age. Explicit user actions (Create PR, Discard, Reset to Draft, deleting
the job or plan, `tendril plan cleanup --force`) call `RemoveWorktrees` directly and are unchanged.

## API Changes

None. `StaleReapStates` is `private static readonly`, `ResolveGrace` is `internal` and keeps its
signature, and no constant or constructor parameter was added or removed. This is a pure behavior
change to the automatic reaper's state policy.

## Files Modified

**Service**
- `src/Ivy.Tendril/Services/Git/WorktreeCleanupService.cs` — dropped `nameof(PlanStatus.Review)`
  from `StaleReapStates`; replaced the comment above it with an explicit statement of why
  `Failed`/`Draft` are reapable and `Review` is not; updated the `ResolveGrace` XML doc to name
  `Review` among the states that are never reaped.

**Tests**
- `src/Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs` — added
  `CleanupPlanWorktrees_NeverReapsReview_EvenFarPastTheStaleWindow` (60-day-idle `Review` plan keeps
  its worktree) and `ResolveGrace_ReturnsNull_ForReview`; moved `Review` out of the two stale-reap
  test arrays and into `RunCleanup_NeverReaps_ActiveStates_EvenWhenOld`; dropped the stale
  `Review` assertion from `ResolveGrace_MapsStatesToWindows`.

## Manual Testing

The reaper is a 30-minute background timer with a 7-day window, so it is not practical to observe
end to end by waiting. To verify by hand:

1. **Confirm the exemption directly.** Run the scoped suite in the worktree:
   `dotnet test src/Ivy.Tendril.Test/Ivy.Tendril.Test.csproj --filter "FullyQualifiedName~Ivy.Tendril.Test.WorktreeCleanupServiceTests"`.
   `CleanupPlanWorktrees_NeverReapsReview_EvenFarPastTheStaleWindow` ages a `Review` plan 60 days and
   asserts its `Worktrees` directory survives; it fails on the pre-change code.
2. **Check the two live at-risk plans.** Plans `00083-InteractiveQuestionBlocksInDraftMarkdown`
   (9 unpushed commits) and `00084-ShowTokenAndCostBreakdownSheetFromJobCostCells` (5) sit in
   `Review` with worktrees present and were last updated 2026-08-14. With this change deployed,
   `D:\Plans\<folder>\Worktrees\` and their `tendril/000NN-…` branches should still exist after
   2026-08-21 (the old 7-day deadline). Before this change they would have been reaped that day.
3. **Confirm explicit removal still works.** Open a `Review` plan in the Review app and press
   Discard. Its `Worktrees` directory should disappear promptly — Discard bypasses the grace check,
   so the exemption must not make a `Review` worktree undeletable.

Note the accepted tradeoff: `Review` worktrees now accumulate on disk until a human acts on the
plan. That is intended — finished work is never destroyed to reclaim disk space.

---

## Fix: Rename RunCleanup_NeverReaps_ActiveStates_EvenWhenOld

The test now asserts over `{ Creating, Executing, Updating, Blocked, Review }`, but `Review` is not an
active/transient state — it is exempt because it holds finished, unpushed commits. Renamed to
`RunCleanup_NeverReaps_UnreapableStates_EvenWhenOld`, which names what the set actually has in common
(no idle age makes them reapable), and rewrote the comment to state both reasons rather than fold
`Review` into "active". No behavior change.

### Files Modified

- `src/Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs` — renamed the test and updated the
  cross-reference to it in `RunCleanup_Skips_Active_State_Plans`.

## Fix: Stop the reaper force-deleting branches that hold unpushed commits

Exempting `Review` closed one state; `Failed` and `Draft` are still reaped after 7 idle days (and
`Completed`/`Skipped`/`Icebox` after 10 minutes), and `RemoveWorktrees` ran an unconditional
`git branch -D` on the plan branch immediately after removing the worktree. A `Failed` plan whose
verifications broke *after* it produced good commits therefore lost them exactly the way plan 00070
did, unattended and with no warning.

`RemoveWorktrees` now takes a `BranchDeleteMode`. The automatic reaper passes `PreserveUnpushed`,
which tries git's safe delete (`git branch -d`) and falls back to `-D` only when a remote-tracking ref
contains the branch tip — safe delete also refuses a *pushed* branch that has no upstream configured,
so that second check prevents the conservative path from hoarding branches whose commits are already
on a remote. When neither holds, the branch is kept and logged with its tip and a recovery hint; the
worktree directory is reclaimed either way, so the cost is a 41-byte ref.

Explicit user actions — Complete, Discard, Reset to Draft, deleting the plan or its ExecutePlan job —
keep the default `Force` mode. Those are deliberate, attended requests to remove the work.

### Files Modified

- `src/Ivy.Tendril/Services/Git/WorktreeCleanupService.cs` — added the `BranchDeleteMode` enum and
  `DeletePlanBranch`/`ForceDeleteBranch`; `RemoveWorktrees` takes the mode (defaulting to `Force`);
  `CleanupPlanWorktrees` passes `PreserveUnpushed`.
- `src/Ivy.Tendril.Test/WorktreeCleanupServiceTests.cs` — four tests over real git repos and
  worktrees: a branch holding unpushed commits survives `PreserveUnpushed` (worktree still gone), a
  branch merged into its base is still deleted, `Force` still deletes regardless, and the same
  guarantee end to end through `CleanupPlanWorktrees` for a stale `Failed` plan. Teardown now clears
  read-only attributes, which git leaves on objects under `.git`.

## Fix: Report a missing worktree and unreachable commits on the Git tab

The symptom that started this plan: for a plan whose worktree is gone, the Git tab rendered a bare
**Worktrees** heading over empty space followed by an unexplained **Commits** table, with nothing
saying the listed commits were reachable from no ref and one `git gc` from destruction.

- An empty worktree list is now an explicit empty state that says why, chosen by plan state:
  removal after a final state, versus the stale reaper reclaiming a plan that sat idle.
- New `GitService.GetCommitRefStatus` classifies a commit as `Reachable`, `Unreachable` (the object
  exists but no ref holds it) or `Missing`. This is a different question from the existing
  `GetReachableCommits`, which asks whether a commit is an ancestor of one worktree's HEAD.
- `GitTabDataBuilder` resolves that for the commits no surviving worktree accounts for, across the
  plan's repos, keeping the answer from the repo that actually has the commit.
- The tab badges those rows (`unreachable` / `not found`) and adds a callout naming the short hashes
  and the `git branch recover/<name> <hash>` escape hatch.

When git cannot answer — an old git without `for-each-ref --contains`, a timeout, an unreadable repo
— the commit is reported `Reachable` and the tab looks exactly as it does today. A false "your work
is about to be destroyed" is worse than no warning.

### Files Modified

**Service**
- `src/Ivy.Tendril/Services/Git/GitResult.cs` — new `CommitRefStatus` enum.
- `src/Ivy.Tendril/Services/Git/IGitService.cs`, `GitService.cs` — new `GetCommitRefStatus`.

**UI**
- `src/Ivy.Tendril/Helpers/GitTabDataBuilder.cs` — `GitTabData.UnassociatedCommitRefStatus` plus the
  per-repo resolution that fills it.
- `src/Ivy.Tendril/Apps/Views/Tabs/GitTabView.cs` — `RenderNoWorktreesState`,
  `BuildUnreachableCommitsCallout`, and the badge in the commit-hash column.

**Tests**
- `src/Ivy.Tendril.Test/Services/GitServiceTests.cs` — real-git coverage of all three statuses,
  including a commit orphaned by deleting its branch.
- `src/Ivy.Tendril.Test/GitTabDataBuilderTests.cs` — status is reported for unassociated commits, the
  repo that has the commit wins in a multi-repo plan, and nothing is claimed when no repo can answer.
- `src/Ivy.Tendril.Test/PlanContentHelpersTests.cs` — fake `IGitService` implements the new member.

### Manual Testing

The user-facing part of this retry is the Git tab, so it is worth looking at:

1. **Empty state.** Open a plan whose worktree is gone but which has commits — plan
   `00070-AddQRCodeGeneratorTool` is exactly this case. The **Worktrees** heading is now followed by a
   card reading *"No worktrees — removed after the plan reached its final state"* (or the stale-reaper
   wording for a `Failed`/`Draft` plan) instead of empty space.
2. **Unreachable badges.** On that same plan, the four commits should each carry a red `unreachable`
   badge next to the hash, with a **Commits At Risk** callout above the table naming them — unless the
   branch has since been restored by hand, in which case they render unbadged as before.
3. **No false alarms.** Open any plan with a live worktree. Its commits sit under the worktree section
   and are unbadged, and no callout appears.

---
## Commits
- 4f6bee67 Report a missing worktree and unreachable commits on the Git tab
- 0c13e77a Stop the reaper force-deleting branches that hold unpushed commits
- e8c8d449 Rename RunCleanup_NeverReaps_ActiveStates_EvenWhenOld to _UnreapableStates_
- 70c4991e Never automatically reap Review plan worktrees

---

### DIFF
--- a/src/Ivy.Tendril.Test/GitTabDataBuilderTests.cs
+++ b/src/Ivy.Tendril.Test/GitTabDataBuilderTests.cs
@@ -14,11 +14,11 @@ public class GitTabDataBuilderTests : IDisposable
             Directory.Delete(_tempDir, recursive: true);
     }
 
-    private static PlanFile CreatePlan(string folderPath, List<string> commits, List<string> prs)
+    private static PlanFile CreatePlan(string folderPath, List<string> commits, List<string> prs, List<string>? repos = null)
     {
         var metadata = new PlanMetadata(
             1, "Test", "Bug", "Test Plan", PlanStatus.Draft,
-            [], commits, prs, [], [], [], DateTime.UtcNow, DateTime.UtcNow, null, null);
+            repos ?? [], commits, prs, [], [], [], DateTime.UtcNow, DateTime.UtcNow, null, null);
         return new PlanFile(metadata, "", folderPath, "");
     }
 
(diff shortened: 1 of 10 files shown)


artem-ivy-ai:
### Tendril staging removed

artem-ivy-ai:
### Staging removed

PR #2163, opened 2026-08-24
Fix stack_hash telemetry, add plan_uuid grouping and three PostHog events

nielsbosma:
## Fix: `stack_hash` was never reaching PostHog

`CreatePlanArgs` is the one plan-job args record with no `PlanFolder` override — the folder doesn't exist when the job is queued — so `TrackSuccessTelemetry` read `""`, `Directory.Exists("")` was always false, and `plan.yaml` was never opened. `stack_hash` was therefore never sent, and `level` was always the hardcoded `"Feature"`.

Now resolved from `job.PlanFile`, which `VerifyCreatePlanResult` records earlier in `HandleCompletion`, with a fallback to the queued project so `stack_hash` still lands when the folder can't be resolved.

**Historical data won't backfill** — existing `plan_created` events lack `stack_hash` and all report `level: Feature`. Filter dashboards to events after this ships.

## New: `plan_uuid` for grouping a plan's events

Plan ids are a per-install sequential counter (`PlanYamlHelper.AllocatePlanId`), so `00042` exists on every install and can't be sent as-is. `plan_uuid` is a SHA-256 of the anonymous id + plan id, stamped with RFC 9562 version 8 / variant bits — globally unique, one-way, and the raw counter never leaves the machine.

Ids are normalized to `D5` first, so the int form from the database (`42`) and the folder-name form (`"00042"`) derive the same UUID. Without that the grouping would silently split.

Added to `plan_created`, `pr_created`, `job_completed`, `plan_state_transition`, and `job_created`.

## New events

| Event | Properties |
|---|---|
| `onboarding_completed` | `project_count`, `agent` |
| `project_created` | `repo_count`, `stack_hash` |
| `job_created` | `job_type`, `agent`, `plan_uuid` |

`project_created` fires from all four in-app creation paths plus the CLI. The CLI runs out-of-process with no DI, so it builds and flushes its own client with a 2s bounded wait — the AddProject promptware invokes it from inside an agent run, where a hung flush would stall the job. `onboarding_completed` flushes immediately (onboarding is often followed by a restart); `job_created` deliberately does not (high volume, `job_completed` flushes the batch).

### ⚠️ `project_created.stack_hash` will be null in practice

`SetupProject` assigns `stackHash` *after* the project exists, via `tendril project set <name> stackHash <hash>`. The property is wired but won't populate today. If the goal is "what stacks do our users run", the event worth adding next is one fired from `ProjectSetCommand` when the field is written.

## Tests

12 new tests in `TelemetryPlanUuidTests.cs` covering UUID stability, int/zero-padded equivalence, distinctness across plans and installs, canonical shape, and null cases.

Full suite: 2329 passed, 1 skipped. (`PromptwareDeployerTests.CreatePrProgram_UsesUniqueTempFileForPrBody` fails only when built to a redirected output directory — it locates its fixture by walking up from `AppContext.BaseDirectory`. Unrelated to these changes.)

Also carries a pre-existing unpushed local commit, `9040aeba "Copy fixes"`.

### DIFF
--- a/src/Ivy.Tendril.Test/Services/TelemetryPlanUuidTests.cs
+++ b/src/Ivy.Tendril.Test/Services/TelemetryPlanUuidTests.cs
@@ -1,0 +1,105 @@
+﻿using Ivy.Tendril.Models;
+using Ivy.Tendril.Services.Telemetry;
+using Xunit;
+
+namespace Ivy.Tendril.Test.Services;
+
+public class TelemetryPlanUuidTests
+{
+    private const string AnonId = "6f1c9c1e-3a5e-4d1a-9d0f-2a6c4b8e1d33";
+    private const string OtherAnonId = "11111111-2222-3333-4444-555555555555";
+
+    [Fact]
+    public void DerivePlanUuid_IsStableForTheSameInstallAndPlan()
+    {
+        var first = TelemetryService.DerivePlanUuid(AnonId, "00042");
+        var second = TelemetryService.DerivePlanUuid(AnonId, "00042");
+
+        Assert.Equal(first, second);
+    }
+
+    [Fact]
+    public void DerivePlanUuid_NormalizesIntAndZeroPaddedForms()
+    {
+        // PlanReaderService supplies the int form from the database; JobCompletionHandler
+        // supplies the zero-padded form parsed off the folder name. Both must group together.
+        Assert.Equal(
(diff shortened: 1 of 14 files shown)


artem-ivy-ai:
### Tendril staging removed

artem-ivy-ai:
### Staging removed

PR #2167, opened 2026-08-26
Ask the user questions from a plan, and answer them in the UI

nielsbosma:
Fenced `questions` blocks end to end: a planning agent asks, the user answers in the UI, and UpdatePlan folds the answers back into the plan.

## What it does

**The schema.** `Prompts/Plans.md` (`## Question Blocks`) is normative — the shape, answer semantics, and lint rules — with `QuestionValidationService` enforcing it from `RevisionWriter.WriteNext`, so the CLI, REST and MCP write paths are all covered. Validation runs before anything touches disk, so a rejected revision does not consume a revision number.

**The picker.** A `questions` fence renders as an interactive block: options, multi-select, free text, a recommended chip. Descriptions are full block markdown, so an option can show the call it proposes rather than describe it. Questions in a block stack rather than tab, and share one Clear.

**Answers are merged, not re-serialized.** `QuestionAnswers.Apply` splices only the `answer` key of the addressed question; comments, key order, scalar style, CRLF and every other block survive byte-for-byte. Reformatting a plan the agent wrote is not a change the user asked for.

**Persistence.** Answers are written back to the *same* revision file — answering is not a new revision, it is filling in a blank the plan left. Manual edits still make a new revision.

**Navigation.** A Questions card indexes the plan, struck through once answered, and clicking an entry scrolls its block into view.

**Read-only.** Plan markdown outside the draft stage — Review, Icebox, the job sheet, Trash — presents the decisions instead of the controls. The plain Markdown widget showed raw YAML.

**Execution.** An unanswered question no longer fails the run. ExecutePlan and RetryPlan resolve one themselves, taking the `recommended` option where there is one, and log the choice. Leaving a question unanswered is a way of saying "you decide", which is why `recommended: true` is load-bearing rather than decoration.

**`optional: true`** marks a question the plan is complete without. It replaces `answer: null`, which had carried that meaning in the wrong place: whether a question must be answered is a property of the question as written, not something done to it afterwards.

## Worth a reviewer's attention

- **Breaking for existing revisions.** `answer: null` is now a lint error with no migration. `D:\Plans` has no organic question blocks, so real exposure looks like zero, but any revision carrying one will fail `write-revision`.
- **`SaveRevision` does not validate.** A hand-edit can persist cross-block duplicate ids; `parseQuestions` dedups per fence and `TryApply` returns on the first match, so answering the second question writes into the first.
- **A write and a full re-parse per keystroke** in free-text fields. Wants debouncing or commit-on-blur; left alone because it changes when answers persist.
- **The read-only swaps have not been checked visually** in Review, Icebox, PlanSheet or Trash.

## Testing

136 C# tests and 123 frontend unit tests over the parser, validator, merge, schema reader and picker. The 11 questions E2E tests passed when last run; assertion changes since then are unverified, because the suite rebuilds the samples project and could not while a local instance held its output.

A code review of the full diff found seven defects, five of which are fixed here — including one that made the feature not work at all: the stale-content guard fired on every answer and reverted it, dropping annotations and letting a second answer overwrite the first on disk.


artem-ivy-ai:

The Tendril staging preview for **PR #2167** (`epic-draft-questions`) has been deleted.

artem-ivy-ai:
### Staging removed

PR #2171, opened 2026-08-28
Remove the Trash feature

nielsbosma:
Removes the Trash feature end to end.

## Why

Trash was never a recycle bin. No UI path ever wrote to it:

- Drafts delete → `PlanReaderService.DeletePlan` removes the folder outright
- Review discard → transitions to `Skipped`
- Icebox delete → same

Its only writer was the CreatePlan promptware via `tendril trash write`, which made it a write-only journal of "the agent declined to create a plan" — expiring after 7 days, and duplicating what the Job Log already records durably.

Usage bore that out: across 34 CreatePlan runs since July, `tendril trash write` executed 5 times, all in one burst caused by the `"New Plan"` placeholder bug, none of them a genuine duplicate rejection.

## The bug it was hiding

Both escape hatches in `VerifyCreatePlanResult` were dead:

- **`IsInTrash` could never match.** `FindTrashEntryById` globs `{planId}-*.md`, but `AllocatedPlanId` is only assigned in `BuildNonCreatePlanFirmware`, so it is always null for a CreatePlan job. The promptware also wrote `<SafeTitle>.md` with no ID prefix.
- **`IsDuplicatePlan` regexed for the literal `identified as duplicate:`**, which no promptware ever emitted.

So a correct duplicate rejection was recorded as **Failed** with *"CreatePlan completed but no plan folder or trash entry was found"*.

## What changed

**Removed:** the Trash app, `tendril trash write`, the sidebar entry and badge, `TrashCount` and its process-lifetime `FileSystemWatcher`, the 7-day cleanup sweep, `FindTrashEntryById`, `IsInTrash`, and `Trash/` directory creation in `ConfigService` / `OnboardingSetupService`. Dropping the trash count left `IConfigService` unused in `TendrilProcessStatusService`, so that constructor parameter went too.

**Replaced:** the promptware is forbidden from writing files directly, so removing the CLI without a replacement would have broken every duplicate rejection. CreatePlan now ends its final message with the `identified as duplicate: <folder>` marker that `IsDuplicatePlan` already looked for. That path now verifies correctly instead of falling through to `MarkCreatePlanFailed`.

The marker match excludes the angle-bracketed template form, so an agent that reads or quotes `Program.md` mid-run cannot echo it and suppress a genuine "no plan produced" failure. `OutputLines` holds re-serialized JSON where `<` arrives escaped as `<`, so both forms are matched — covered by a new test.

The other two decline paths (no project matched, code-state validation failed) report their reason and exit. They were already recorded as Failed before this change, so their behavior is unchanged.

## Notes

- Historical release-note entries still mention Trash. Those releases did ship it; that is left as-is.
- Existing installs keep a stale `TENDRIL_HOME/Trash` folder. Inert markdown, deliberately not deleted on upgrade.

## Verification

- Builds clean: `Ivy.Tendril`, `Ivy.Tendril.Test`, `Ivy.Tendril.Docs`, `Ivy.Tendril.Test.End2End` — 0 warnings
- Full unit suite: **2336 passed, 0 failed, 1 skipped**
- `tendril trash write` → `Unknown command 'trash'`; `tendril --help` clean

https://claude.ai/code/session_01MUZ4r8MqmZ6aGHPyLf7rcr


artem-ivy-ai:
### Staging removed

artem-ivy-ai:
### Tendril staging removed

PR #2196, opened 2026-08-28
Render sanitized <details> blocks in DraftMarkdown

nielsbosma:
## Problem

Tendril's own promptware instructs agents to emit GitHub-style `<details>`/`<summary>` blocks — `Promptwares/UpdatePlan/Program.md` builds the plan `## Questions` section out of them, and `ExecutePlan/Program.md` uses one for "Still relevant?" — but `react-markdown` drops raw HTML unless `rehype-raw` reparses it. Every such block rendered as literal escaped text with the body permanently expanded:

```html
<h2>Questions</h2>
&lt;details&gt;
&lt;summary&gt;Should we use X?&lt;/summary&gt;
<p>Yes, because Y.</p>
&lt;/details&gt;
```

The framework's own `Markdown` widget has had this support all along (`rehype-raw` plus `details`/`summary` component overrides in `MarkdownRenderer.tsx`); DraftMarkdown never did.

## Approach

Parse raw HTML with `rehype-raw`, then immediately prune it with `rehype-sanitize` — the content is model-written and never trusted. The allow-list starts from `hast-util-sanitize`'s GitHub-derived `defaultSchema` (which already permits `details`, `summary` and `open`) with three adjustments in the new `frontend/src/rawHtml.ts`:

| Adjustment | Why |
| --- | --- |
| `strip: [...default, "style"]` | Dropping an element otherwise keeps its children, so a `<style>` block's CSS rules landed on the page as prose. Caught in a live browser pass, not in the unit tests. |
| `code: [["className", /^language-./, "math", "math-display", "math-inline"]]` | `remark-math` emits `<code class="language-math math-display">` and `rehype-katex` finds those nodes by class. The default schema allows only `language-*`, so the marker class was stripped and **all math silently vanished**. |
| `protocols: {}` | react-markdown's `urlTransform` already runs after this pass with the same protocol list. A second one here only strips the `file://` and `D:\...` links that `DangerouslyAllowLocalFiles` exists to render. |

Two ordering rules hold the pipeline together:

- sanitising runs **before** `rehype-katex`, whose output is a large tree of classed spans, inline styles and MathML the allow-list would shred;
- both rehype passes are gated on the content actually containing HTML, matching how math is already gated, so the common case keeps its current render path.

This reaches every markdown surface, not just DraftMarkdown — `AgentViewer`, `ChatWidget` and `PlanDiffView` all build their plugins through `getMarkdownPlugins`.

Styling mirrors the framework's `typography.details` / `typography.summary`: card border, hover, body inset, and a chevron drawn as a CSS mask so it follows `currentColor` in both themes.

## Sample

New `Collapsible` sample at `/draft-markdown/collapsible`, covering basic sections, `open`, rich bodies (list + table + code + alert), nesting, and a sanitisation section that carries live `onclick`, `<script>`, `<iframe>`, `<style>` and `javascript:` payloads so the pruning is visible rather than merely asserted.

## Verification

- `tsc --noEmit` clean
- 130/130 vitest pass, including 13 new cases: rendering, `open`, nesting, each sanitisation class, plus regression guards that math, highlighted code, task lists, tables and local-file links all survive the extra pass
- widget bundle builds; samples project compiles
- ran the sample against a live server: 9 `<details>` elements, 1 open, zero surviving `script` / `iframe` / `style` / `onclick` / `javascript:` hrefs, chevron mask applied, no console errors

**Not run:** the new `collapsible.spec.ts` E2E spec. A leftover samples server was holding a lock on the build output, and stopping someone else's process to run it wasn't worth it — the unit tests cover the same assertions. It will be exercised on the next full Playwright run.

## Note for review

`defaultSchema` is the GitHub-safe tag set, so raw `<b>`, `<kbd>`, `<sub>` etc. now render too. That is GitHub-equivalent and matches the framework's behaviour, but it is broader than `<details>` alone. Narrowing `tagNames` to just details/summary is a one-line change in `rawHtml.ts` if that is preferred.

https://claude.ai/code/session_01Pooaou9FqrDRse6ZBAUdDV

### DIFF
--- a/src/Ivy.Tendril.Widgets/.samples/Apps/DraftMarkdown/CollapsibleApp.cs
+++ b/src/Ivy.Tendril.Widgets/.samples/Apps/DraftMarkdown/CollapsibleApp.cs
@@ -1,0 +1,128 @@
+using Ivy;
+using Ivy.Tendril.Widgets;
+using DraftMarkdownWidget = Ivy.Tendril.Widgets.DraftMarkdown;
+
+namespace WidgetSamples.Apps.DraftMarkdown;
+
+[App(title: "Collapsible", icon: Icons.ListCollapse, group: ["DraftMarkdown"])]
+class CollapsibleApp : ViewBase
+{
+    public override object Build()
+    {
+        var markdown = """
+            # Collapsible Sections
+
+            GitHub-style `<details>` / `<summary>` blocks render as real disclosure
+            widgets. Tendril's own promptware emits them (the plan `## Questions`
+            section), so plan markdown routinely contains raw HTML.
+
+            ## Basic
+
+            <details>
+            <summary>What database does the plan target?</summary>
+
+            PostgreSQL 16. The migration scripts live in `db/migrations` and are
+            applied by the `MigrateDatabase` job on startup.
+
+            </details>
+
+            <details>
+            <summary>Do we need a feature flag?</summary>
+
+            No. The change is additive and the old code path is removed in the same
+            commit, so there is nothing to toggle between.
+
+            </details>
+
+            ## Open by Default
+
+            <details open>
+            <summary>Still relevant?</summary>
+
+            Yes. The `open` attribute survives sanitisation, so a section can start
+            expanded.
+
+            </details>
+
+            ## Rich Bodies
+
+            A section body is ordinary markdown — lists, tables, code and diagrams all
+            work inside one.
+
+            <details>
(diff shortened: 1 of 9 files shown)


artem-ivy-ai:
### Staging removed

artem-ivy-ai:

The Tendril staging preview for **PR #2196** (`feat/draft-markdown-details`) has been deleted.

PR #2265, opened 2026-09-02
Review a running app: page-scoped comments, queued updates, richer change requests

nielsbosma:
Continues the WebViewer epic. A review action already became a live preview of the app it starts; this makes the comments left on that preview actually usable, and fixes what driving it against a real Vite app turned up.

## Comments belong to a page

A pin resolved by xpath or CSS path finds *something* on almost any page, so navigating away did not leave a pin behind visibly — it silently re-attached to whatever occupied that position and looked entirely legitimate. Comments now carry the canonical page they were left on (`pageUrl.ts`: hash dropped, trailing slash dropped, query kept), pins are pushed filtered by it, and the agent re-checks that a resolved node still matches the tag that was picked.

The change request groups under those pages. "Make this green" says nothing without knowing where "this" was.

## Repeated updates queue instead of colliding

Each `RetryPlan` now waits on every unfinished job for the plan, `Blocked` ones included, so a third Update queues behind the second rather than putting a third agent on the same worktree. Sending is refused unless the plan is in `Review` or already applying a retry — checked when the dialog opens and again on the click, since it can sit open across a state change.

## The change request says which element

Before, an agent got a file and a line. A comment on an input resolved to `src/components/ui/input.tsx` — the shared primitive — and acting on that restyles every input in the app when one screen was meant.

Almost all of the fix was plumbing: `describe()` already captured the element's text and identifying attributes and the widget forwarded only the tag, and `DebugJson` already carried `ownerChain`, `provenance` and `confidence` while only file and line were read. Now:

```
- **2. <textarea>** "Enter text to encode or decode..."
  Placeholder is too vague - say what format it expects
  source: `src/components/ui/input.tsx:8` (high, react-owner-stack)
  component: Input > HtmlEncoder > ToolPage > App
  attributes: data-testid="encoder-input"
  viewport: Desktop
```

A preamble explains that the call site, not the primitive, is usually the target, and that anything below high confidence is a lead to verify. The selector prints only when nothing resolved, where it is the last handle left.

## Fixes found by using it

- **Back and forward did nothing.** A client-side route change reports its URL without re-pointing the frame, so `frameSrc` stayed at the first page; setting it back to itself was a no-op React discarded, and the index and address bar moved while the page sat still.
- **Fonts were missing.** The proxy's local-only gate refused `fonts.googleapis.com`, so the app under review rendered in fallback fonts with its icons gone. It now admits a fixed list of static-asset hosts over https — exact host match, no wildcards.
- **Select was one-shot.** It stays on until stopped; Escape cancels the pick, not the mode, so the page's flag and the toolbar cannot drift apart.
- **Toolbar.** Icon device switcher, address bar capped so it stops swallowing the row, controls matching the location bar's height.
- **`ObjectDisposedException` on shutdown.** `PlansChanged` is raised on a timer thread against an invocation list snapshotted before a view could unsubscribe, so a disposed coalescer could still be called. A late `Request()` is now ignored rather than thrown at — it was being logged as "PlansChanged subscriber threw", which is a net meant for real bugs.

## Verification

Build clean. 149 C# tests on the `AppPreview|WebViewer|ReviewAction` filter plus the coalescer's 5; `tsc --noEmit` clean and 263 vitest tests across 25 files. Not the full 2400-test suite.

Not verified in a browser: the toolbar sizing and the tablet icon are unreviewed visually.

## Known gap

A project-scoped review action (added on development while this branch sat) has no plan, so it keeps the terminal and never becomes a preview — the preview's whole output is a change request against a plan. Closing it means making the view take a nullable plan and hiding Update. Left as a follow-up rather than folded into the merge.

https://claude.ai/code/session_01WkJwTMwcAj7dA9P8gJM5T9


artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-tendril-staging-docs-2265.sliplane.app](https://ivy-tendril-staging-docs-2265.sliplane.app)

artem-ivy-ai:
### 🚀 Tendril Staging Preview Ready

- **Branch:** `epic-webviewer` (PR #2265)
- **Environment URL:** [https://ivy-tendril-tendril-pr-2265.sliplane.app](https://ivy-tendril-tendril-pr-2265.sliplane.app)

artem-ivy-ai:
### Staging preview

**📄 Docs:** [https://ivy-tendril-staging-docs-2265.sliplane.app](https://ivy-tendril-staging-docs-2265.sliplane.app)

artem-ivy-ai:
### 🚀 Tendril Staging Preview Ready

- **Branch:** `epic-webviewer` (PR #2265)
- **Environment URL:** [https://ivy-tendril-tendril-pr-2265.sliplane.app](https://ivy-tendril-tendril-pr-2265.sliplane.app)

PR #2326, opened 2026-09-05
[00100] Make Jobs Table Headers Match Their Filter Column Names

rorychatt:
## Problem

The Jobs table has filtering enabled, and Ivy's filter grammar addresses a column by its **row property name**, never by the header the user reads. Three of the ten headers do not name the property they are bound to, so what a user reads is not what a user can type:

| Header today | Token that actually works | What happens |
| --- | --- | --- |
| Plan | `[PlanId]` | `[Plan]` is valid but filters the prompt column instead, silently |
| Prompt/Title | `[Plan]` | not guessable, and `/` is not part of any token |
| Status (the second one) | `[StatusMessage]` | two columns both read "Status" |

The failure mode is not an error, it is a wrong answer: `[Plan] = "00100"` parses, runs against the prompt text, and returns zero rows with nothing to say the wrong column was filtered.

## Solution

Make the headers match the property names by:

1. **Rename `JobItemRow.Plan` to `Prompt`** - Its value is the prompt display text, so `Prompt` is the accurate name
2. **Delete the ten `.Header()` calls** - Let Ivy derive headers from property names (e.g., `PlanId` → "Plan Id", `StatusMessage` → "Status Message")
3. **Take hidden columns out of the filter list** - Add `.Filterable(t => t.Id, false)` etc. for `Id`, `LastOutputTimestamp`, and `ErrorContext`

After this change:
- "Plan" becomes "Plan Id" (filter: `[PlanId]`)
- "Prompt/Title" becomes "Prompt" (filter: `[Prompt]`)
- Duplicate "Status" becomes "Status Message" (filter: `[StatusMessage]`)

---

**Commits:**
- 534e9c3f

---

### DIFF
--- a/src/Ivy.Tendril.Test/Apps/Jobs/JobsTableColumnLabelTests.cs
+++ b/src/Ivy.Tendril.Test/Apps/Jobs/JobsTableColumnLabelTests.cs
@@ -1,0 +1,70 @@
+using Ivy.Tendril.Models;
+using Xunit;
+
+namespace Ivy.Tendril.Test.Apps.Jobs;
+
+public class JobsTableColumnLabelTests
+{
+    private static readonly Dictionary<string, string> ExpectedLabels = new()
+    {
+        { nameof(JobItemRow.Id), "Id" },
+        { nameof(JobItemRow.Status), "Status" },
+        { nameof(JobItemRow.PlanId), "Plan Id" },
+        { nameof(JobItemRow.Prompt), "Prompt" },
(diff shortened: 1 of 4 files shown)


artem-ivy-ai:
### Staging removed

artem-ivy-ai:

The Tendril staging preview for **PR #2326** (`tendril/00100-MakeJobsTableHeadersMatchTheirFilterColumnNames`) has been deleted.

PR #2337, opened 2026-09-05
[00147] Fix Flaky Hanging Hook Test by Making Hook Guards Injectable

rorychatt:
## Problem

JobServiceHookTests.RunHooks_HangingBeforeHook_IsKilledAndDoesNotWedgeLaunch failed once in four full-suite runs during execution. It spawns PowerShell running Start-Sleep -Seconds 120, expects a 10 second guard to kill it, and allows 30 seconds for the whole thing.

The test has a pass margin of about 3x, but nothing in the test is cheap enough to absorb a loaded machine. Two load-sensitive costs sit inside that margin:

1. **pwsh startup** - The bundled PowerShell startup can take seconds under contention or antivirus scanning
2. **Thread pool scheduling** - Before-hooks run synchronously inside StartJob, so Task.Run parks a pool thread on a blocking WaitForExit. When the pool is saturated by other suites, the work item may not start for seconds while the 30s allowance is already counting.

## Solution

Make the two hook guards injectable as internal test seam properties on JobService, matching the existing JobTimeout and StaleOutputTimeout. The hanging test then injects a 1s condition guard, which drops its expected runtime from 10s to about 1s while keeping the 30s allowance unchanged (roughly 30x headroom instead of 3x).

### Changes:
1. **JobCompletionHandler** - Replace hard coded guards with injectable TimeSpan properties (HookConditionTimeout, HookActionTimeout)
2. **JobService** - Expose the guards to the test project via pass-through properties
3. **JobServiceHookTests** - Inject 1s guard in the hanging test and replace Task.Run with dedicated thread to avoid pool contention
4. **New test** - HookGuards_DefaultToProductionValues pins the production defaults (10s condition, 30s action)

---

**Commits:**
- 6a30990d Make hook guards injectable to fix flaky test

---

### DIFF
--- a/src/Ivy.Tendril.Test/JobServiceHookTests.cs
+++ b/src/Ivy.Tendril.Test/JobServiceHookTests.cs
@@ -354,7 +354,9 @@ public class JobServiceHookTests : IDisposable
     // Regression for the #1455 class: a before-hook runs synchronously in the un-monitored pre-launch
     // window (before JobMonitor/TimeoutCts exist), so a hook that never returns would wedge the launch
     // forever with no timeout able to fire. The hook process must be force-killed at its guard
-    // (10s for conditions, 30s for actions) rather than blocking on a ReadToEnd that never completes.
+    // (defaults: 10s for conditions, 30s for actions) rather than blocking on a ReadToEnd that never
+    // completes. This test overrides the condition guard to 1s so the test runtime does not track
+    // pwsh startup and thread pool contention on a loaded machine.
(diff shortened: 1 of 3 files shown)


artem-ivy-ai:
### Staging removed

artem-ivy-ai:

The Tendril staging preview for **PR #2337** (`tendril/00147-FixFlakyHangingHookTestByMakingHookGuardsInjectable`) has been deleted.

PR #2390, opened 2026-09-07
[00055] Replace Dashboard Trend Average Line With Rolling 7 Day Average

rorychatt:
# Plan 00055 — Replace Dashboard Trend Average Line With Rolling 7 Day Average

## What changed

The dashboard trend card drew one horizontal dashed line at the mean of whatever was on screen. It
is now a smoothed curve of the 7-day trailing average, and to make that mean anything the underlying
series had to become daily.

**Both ranges are daily.** The long range plotted 12 monthly buckets and the short range fell back to
4 weekly buckets when the daily series was absent. Both paths are gone. `BuildDailyTrend` emits one
point per calendar day — 365 for the long range, 28 for the short one — with `yyyy-MM-dd` dates
instead of month or day labels. A rolling mean over buckets would have been a "7 buckets" average
wearing a "7-day average" label, so when no daily series exists the builders return null and the card
does not render at all.

**Quiet days now count against the average.** Days with no rows are zero-filled, and the calculator
always divides by 7 rather than by the days that happened to have data, so a week with two idle days
reads lower than a week without them. That is only honest if a zero can be distinguished from an
absence, which is what `DashboardActivityStats.DailyDataStart` is for: the earliest day any plan or
cost row exists, clamped up to the retrieval window. Days on or after it with no rows are genuinely
zero; the rolling value is null until the whole window sits inside recorded history, and a null is a
gap in the curve rather than a number dragged down by days nobody recorded.

**The series reaches further back.** `DailyCostWindowDays = 60` became
`DailyTrendWindowDays = 736`: 365 displayed days, six leading days so the first plotted point has a
full window, and 365 more for the prior-year comparison. The forecast re-applies its own 30 day
cutoff, so it cannot move — the ten `CostForecastCalculatorTests` are in scope for exactly that
reason and all pass.

**The frontend plots it and owns the calendar.** `TrendChart` takes `dates` and `rolling` instead of
`labels` and a scalar `average`, splits the rolling series on nulls into one `.tdb-trend-avg-curve`
path per contiguous run, and draws it over the comparison and under the current line. The rolling
values count towards the y-scale, because the first points average days from before the displayed
range and can sit above everything on screen. Axis ticks and tooltip titles come from the dates,
parsed by hand: `new Date("2026-01-01")` is UTC midnight and renders as December 31 in every
negative-offset time zone.

**The legend is present even when the curve is not.** A fresh install with under seven days of
history sees a muted `7-day average (needs 7 days of history)` rather than a silently missing item.

## Decisions taken

Both question blocks were unanswered, so the recommended option was taken and logged to job 00292:

- **`long-range-span` → `twelve-months-daily`.** 365 daily points, the `Last 12 months` label and the
  prior-year comparison unchanged, 736 day retrieval window. The two alternatives (90 or 30 days)
  would have changed what the toggle means and lost the year-over-year read.
- **`comparison-smoothing` → `raw-comparison`.** The comparison series stays raw daily values, as the
  dashed line does today; only the current series gains a curve.

## Commits

| Commit | Subject |
| --- | --- |
| `40cfcd46` | Add a rolling average calculator and a daily data start |
| `9a9731dd` | Build both trend ranges as daily series with a rolling average |
| `eb06ea56` | Replace the trend chart's average line with a rolling curve |

18 files, +1021 / −212.

## Verification

| Check | Result |
| --- | --- |
| DotnetFormat | Pass — `--verify-no-changes` clean on all ten changed C# files |
| DotnetBuild | Pass — every project the solution owns, `--warnaserror --no-incremental`, 0 warnings |
| DotnetTest | Pass — 60 of 60 in the plan's filter |
| CheckResult | Pass — all eight solution sections implemented as written |
| NpmTest | Pass — 45 of 45 across five files in `src/TendrilDashboard` |
| NpmBuild | Pass — `tsc` clean, bundle built |

Two pre-existing failures were reproduced at the base commit and are unrelated to the dashboard: the
`Ivy.Tendril.Docs` project cannot compile against the published `Ivy.Docs` package (it needs
`Ivy.Docs.Helpers`, which only exists in the framework source), and three `PromptwareDeployerTests`
assert Debug-build behaviour while the verification runs Release. Details in `Verification/`.

`src/Ivy.Tendril/Ivy.Tendril.slnx` lists `../../../Ivy-Framework/src/Ivy/Ivy.csproj`, which resolves
beside the original repository but not beside a worktree, so the solution cannot restore here at all.
Project reference paths were left untouched and no filesystem aliases were created; every project was
built and tested individually instead, where `src/Directory.Build.props` resolves `IvySource=false`
and Ivy comes from its published packages.

---

### DIFF
--- a/src/Ivy.Tendril.Test/Apps/DashboardAppViewModelTests.cs
+++ b/src/Ivy.Tendril.Test/Apps/DashboardAppViewModelTests.cs
@@ -49,48 +49,118 @@ public class DashboardAppViewModelTests
         Assert.Null(kpi.Direction);
     }
 
+    /// <summary>A daily cost row per day in the inclusive range, priced by <paramref name="cost" />.</summary>
+    private static List<DashboardDailyCost> DailyCosts(DateOnly from, DateOnly to, Func<DateOnly, decimal> cost) =>
+        Enumerable.Range(0, to.DayNumber - from.DayNumber + 1)
+            .Select(i => from.AddDays(i))
+            .Select(day => new DashboardDailyCost(day, cost(day), 0))
+            .ToList();
+
+    /// <summary>Period 11 against a window of 7, so the rolling mean cannot come out flat.</summary>
+    private static decimal Sawtooth(DateOnly date) => (date.DayNumber % 11) * 10m;
+
+    /// <summary>A different value on every day, so an assertion can pin which day was read.</summary>
+    private static decimal Unique(DateOnly date) => date.DayNumber;
+
     [Fact]
-    public void BuildTrend_TakesTrailingTwelveMonths()
+    public void BuildTrend_ProjectsAYearOfDailyPoints()
     {
-        var months = new List<DashboardMonthStats>();
-        for (var month = 1; month <= 12; month++)
-            months.Add(new DashboardMonthStats(2025, month, month, 0, month * 100m, 0));
-        for (var month = 1; month <= 8; month++)
-            months.Add(new DashboardMonthStats(2026, month, month * 2, 0, month * 200m, 0));
+        var today = new DateTime(2026, 9, 6);
+        var dataStart = new DateOnly(2024, 9, 1);
+        var activity = new DashboardActivityStats(
+            [], 0m, DailyCosts(dataStart, new DateOnly(2026, 9, 6), Sawtooth), null, null, dataStart);
 
-        var trend = DashboardApp.BuildTrend(new DashboardActivityStats(months, 0));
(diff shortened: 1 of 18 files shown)


artem-ivy-ai:
### Staging removed

artem-ivy-ai:

The Tendril staging preview for **PR #2390** (`tendril/00055-ReplaceDashboardTrendAverageLineWithRolling7DayAverage`) has been deleted.

